PR #13, opened 2024-03-25
Alice

Svannia:
The whole user creation, sign in feature and general structure of the backbone.
That's a big one haha sorry.
I'm doing a PR now since the code seems to be working, if I pushed less there would be bugs.


Ovoland (approved):
The restructuration and the new additions seem good to me.

Ovoland:
The CI fails, we should find a way to solve this

PR #14, opened 2024-03-25
Alice

Svannia:
I have added all the correct navigation to account creation, homescreen and two different settings menu, with a drawer menu.
The user can see the groups they are in, but I havent made the group creation feature yet.

Edit: a group can now be created properly with the latest commit. Also added some more navigation to access more menus


app/src/main/java/com/github/se/studybuddies/viewModels/UserViewModel.kt
```diff
@@ -12,53 +12,54 @@ import kotlinx.coroutines.launch
...
+
+  fun getCurrentUserUID(): String {
+    return db.getCurrentUserUID()
+  }
+
+  fun fetchUserData(uid: String) {
+    viewModelScope.launch {
+      try {
+        val document = db.getUserData(uid).await()
+        if (document.exists()) {
```
david-schroeter, line 37: Same as in `GroupViewModel`, I think extraction should be performed directly in the `DatabaseConnection` class. What do you think ?

app/src/main/java/com/github/se/studybuddies/ui/settings/CreateAccount.kt
```diff
@@ -17,59 +17,54 @@ import androidx.compose.runtime.mutableStateOf
...
-    val photoState = remember { mutableStateOf(Uri.EMPTY) }
+  val email = FirebaseAuth.getInstance().currentUser?.email ?: ""
+  val usernameState = remember { mutableStateOf("") }
+  val photoState = remember { mutableStateOf(Uri.EMPTY) }
 
-    LaunchedEffect(key1 = true) {
-        val defaultProfilePictureUri = withContext(Dispatchers.IO) { userViewModel.getDefaultProfilePicture() }
-        photoState.value = defaultProfilePictureUri
-    }
+  LaunchedEffect(key1 = true) {
```
david-schroeter, line 37: In my opinion the `key1` argument should not always be equal to `true`. The purpose of this argument is to determine whether or not to cancel and restart the coroutine upon recomposition. So if it is set to `true` it will never be restarted upon recomposition.

[Here](https://developer.android.com/develop/ui/compose/side-effects#restarting-effects) is a most detailed explanation of the `key` argument in LaunchEffect function.

How about giving `uid` as the value of `key1` argument, what do you think ?

app/src/main/java/com/github/se/studybuddies/ui/settings/CreateAccount.kt
```diff
@@ -17,59 +17,54 @@ import androidx.compose.runtime.mutableStateOf
...
+  Column(modifier = Modifier.fillMaxSize()) {
+    LazyColumn(
+        modifier = Modifier.fillMaxWidth().padding(horizontal = 20.dp, vertical = 20.dp),
+        verticalArrangement = Arrangement.spacedBy(20.dp, Alignment.Top),
+        horizontalAlignment = Alignment.CenterHorizontally) {
+          item {
+            Column(
+                modifier = Modifier.fillMaxWidth(),
+                verticalArrangement = Arrangement.spacedBy(20.dp)) {
+                  Text("You signed in with the email address $email")
```
david-schroeter, line 57: It is a good practice to extract all hard coded string that are displayed to the user inside the strings.xml resource file. That way if some day you decide to localize your app and provide a French version for example, you only need to create a french version of the strings.xml file without having to navigate the whole code base to extract each string to this file.

Here is the link to the [localization page](https://developer.android.com/guide/topics/resources/localization#resource-precedence) in the android documentation.

app/src/main/java/com/github/se/studybuddies/ui/groups/CreateGroup.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.github.se.studybuddies.viewModels.GroupViewModel
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+@Composable
+fun CreateGroup(groupViewModel: GroupViewModel, navigationActions: NavigationActions) {
+  val nameState = remember { mutableStateOf("") }
+  val photoState = remember { mutableStateOf(Uri.EMPTY) }
+
+  LaunchedEffect(key1 = true) {
```
david-schroeter, line 33: Same as in `CreateAccount`, should `key1` really be set to `true`?

david-schroeter (commented):
Nice job on this PR, the newly added features look really nice!

I've added a couple of comments, while reviewing your PR that I think should be addressed.

Some comments that I wrote once actually appeared multiple times over the code, I am especially thinking about the hard coded UI strings that should be extracted in a `String.xml` file.

As correctly mentioned by @cemuelle, there is currently no test of the new features. It is very important that newly created features get tested alongside their creation to avoid increase in the technical debt, and improve the code maintainability.

Once you have implemented the feature, the test is the part that checks that what you have implemented actually does what you expect it to do.

So I would suggest you to add those tests before merging this PR.

But very good work so far on this PR, nicely done! 👍🏻

app/src/main/java/com/github/se/studybuddies/viewModels/DatabaseConnection.kt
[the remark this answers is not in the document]
Svannia: Yes, that seems like a good idea. I was actually hesitating about where to place this file so I'll move it how you suggested.

app/src/main/java/com/github/se/studybuddies/viewModels/DatabaseConnection.kt
```diff
@@ -2,104 +2,179 @@ package com.github.se.studybuddies.viewModels
...
+  }
 
-    fun getUserData(uid: String): Task<DocumentSnapshot> {
-        return userDataCollection.document(uid).get()
-    }
-    suspend fun getDefaultProfilePicture(): Uri {
-        return storage.child("userData/default.jpg").downloadUrl.await()
-    }
+  fun getUserData(uid: String): Task<DocumentSnapshot> {
+    return userDataCollection.document(uid).get()
```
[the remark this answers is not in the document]
Svannia: I coded this way because this is what a TA showed me during the bootcamp. But other teammates would also prefer the DatabaseConnection to return the data struct. However this also changes a lot how I coded the viewModels, so I will be fixing that nice and clean next sprint.

app/src/main/java/com/github/se/studybuddies/viewModels/DatabaseConnection.kt
```diff
@@ -2,104 +2,179 @@ package com.github.se.studybuddies.viewModels
...
-                    }
-                    .addOnFailureListener { e ->
-                        Log.d("MyPrint", "Failed to upload photo with error: ", e)
-                    }
-                Log.d("MyPrint", "User data successfully updated")
-            }
-            .addOnFailureListener { e ->
-                Log.d("MyPrint", "Failed to update user data with error: ", e)
-            }
+  fun createUser(uid: String, email: String, username: String, profilePictureUri: Uri) {
```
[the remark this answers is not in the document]
Svannia: Same as the next comment, I will clean all database related data retrieval and creation next sprint. Thanks for the feedback!

app/src/main/java/com/github/se/studybuddies/ui/settings/CreateAccount.kt
Svannia: That is great, I didn't know about this! Thanks for the tip.

app/src/main/java/com/github/se/studybuddies/ui/groups/CreateGroup.kt
Svannia: I must admit I kind of wrote this blindly (code from ChatGPT) because I didn't really understand it. This may be why I had some issues with the pages not loading.

Ovoland:
This is a really good job and I agree with the comments of the TA. However as already solved before this pull request is a bit too big. Also the tests aren't checking everything (for example the verification of the sign in functionality and the group creation (can you save if no name are entered)). Plus some comments from the TAs aren't answered yet (launched effet = key1). However due to the lack of time before the deadline and the urge to handle the other branches that are related to this one, I will accept this pull request knowing that it's not perfectly fonctional. We'll have to come again to these features and fix all the remaing issues.  But otherwise, good job !

app/src/androidTest/java/com/github/se/studybuddies/tests/LoginTest.kt
```diff
@@ -0,0 +1,44 @@
...
+  // is deprecated, but it was MUCH faster than using IntentsRule in our tests
+  @get:Rule val intentsTestRule = IntentsTestRule(MainActivity::class.java)
+
+  @Test
+  fun titleAndButtonAreCorrectlyDisplayed() {
+    ComposeScreen.onComposeScreen<com.github.se.studybuddies.screens.LoginScreen>(composeTestRule) {
+      // Test the UI elements
+      loginTitle {
+        assertIsDisplayed()
+        assertTextEquals("Welcome")
```
Ovoland, line 36: This should be changed by the actual name of the app

Ovoland (approved):
Already commented the pull request (check the previous comment)

PR #34, opened 2024-04-08
UI reformating

Ovoland:
I change the UI visual of the Sign In page, the group page and the global visual. I also implemented to top navigation bar but haven't finished it yet. This first pull is to test whether the the code passes the CI


TdVBd (commented):
Overall the code is clear and functional. 
The way the navigation bar is related to the drawer menu is a bit unintuitive and could appear impractical if in the future we wish to dissociate these two UI features, but for now we plan to always use them together so it should not cause any trouble.
Also, the way the scafold, lazy collumns and collumns structures are used does not seem to follow a very coherent pattern.

TdVBd (approved):
Overall the code is clear and functional.
The way the navigation bar is related to the drawer menu is a bit unintuitive and could appear impractical if in the future we wish to dissociate these two UI features, but for now we plan to always use them together so it should not cause any trouble.
Also, the way the scafold, lazy collumns and collumns structures are used does not seem to follow a very coherent pattern.

PR #61, opened 2024-04-12
Loading screen

oliver216:
Add a loading screen when waiting for the database to load the groups. This make the app more smooth.
This loading screen have been implemented so it can easily be used to other place later.


Ovoland:
Good addition. This is a bit like a "fake resolution" but it actully does the job !

PR #63, opened 2024-04-12
Chat Screen

cemuelle:
Add a chat screen where multiple users can send and receive messages


Freezemoon (approved):
Nice code! The tests are a good addition and they all pass! Very good modularization too. Maybe next PR try to do a smaller amount of lines, this one is quite big. There are some ToDos left but mostly about small bugs that should be debugged later but don't really impact the app or now. Good job!

PR #72, opened 2024-04-17
Permission request for group profile picture

Freezemoon:


Svannia:
Looks good to me, as said in-person, see if you can add support (new permission READ_MEDIA) for Android 14.0, API/SDK level 34 so that it also works with the most updated devices.

Svannia (approved):
Looks all good to me with the new permission support!

PR #73, opened 2024-04-17
Profile permissions

Freezemoon:
Similar to the PR for permissions in Group creation/modification. Also coded an additional utilitary function that can launch the permission requests of multiple permissions one after the other if needed in the future.


app/src/main/java/com/github/se/studybuddies/ui/permissions/MultiplePermissions.kt
cemuelle, line 1: It seem's to have some duplicate lines between checkMultiplePermissions and  checkPermission. I suggest to modify checkPermission to call checkMultiplePermissions with as argument a list of only one element.

app/src/main/java/com/github/se/studybuddies/ui/settings/AccountSettings.kt
cemuelle, line 1: As for CreateGroup, you have getContent.launch("image/*") with always the same text. I suggest to extract it and maybe have a single variable for both files.

app/src/main/java/com/github/se/studybuddies/ui/settings/CreateAccount.kt
cemuelle, line 1: Still the same comment as for AccountSettings and CreateGroup concerning the "image/*"

cemuelle (changes requested):
Great job overall, just the "image/*" text that should be extracted and put in a constant.
And also maybe modify checkPermission to avoid duplicate code with checkMultiplePermissions.

cemuelle (approved):
The suggestions have been applied

PR #75, opened 2024-04-19
add path to the test for sonar

cemuelle:
It seem's that Sonar doesn't find Jacoco test report, so I added new properties to indicates the path


cemuelle:
Closed because it's seems working

PR #84, opened 2024-04-24
Add features and fix issues for the chat

cemuelle:
It contains multiple new features for the chat as :

- Delete messages
- Edit messages
- Display time of send of a message
- Display user profile picture next to each message
- Add automatical scroll down
- Add group title, with group picture, name and members name

Fix some issues like when open keyboard, the chat was going up.

And also extract some part to create own composable to simplify the code
Also modify some part to merge them as one with parameters to choose which version we want


app/src/main/java/com/github/se/studybuddies/data/Message.kt
Freezemoon, line 1: Maybe we can do this later on but I thought it would be more accessible if the format of dates and hours was specific to each person's. Which means instad of one hardcoded format, we would directly take the format of the actual phone.

app/src/main/java/com/github/se/studybuddies/ui/ChatScreen.kt
Freezemoon, line 1: Commented out line about the profile picture, not sure if its for later ?
Also, don't forget to put hardcoded strings into the string.xml file (mostly for the content descriptions)

app/src/main/java/com/github/se/studybuddies/viewModels/MessageViewModel.kt
Freezemoon, line 1: Im not sure the list of random users if necessary ?

app/src/main/java/com/github/se/studybuddies/ui/SharedUI.kt
Freezemoon, line 1: Another string that should be extracted

app/src/main/java/com/github/se/studybuddies/data/Message.kt
Freezemoon, line 1: Maybe we can do this later on but I thought it would be more accessible if the format of dates and hours was specific to each person's. Which means instad of one hardcoded format, we would directly take the format of the actual phone. Maybe keep in mind for V2

Freezemoon (changes requested):
Overall good job ! Nice to see you managed to resolve some bugs that were present in the earlier version and the added features are very nice ! Just some nitpicking about hardcoded strings and some minor things to change or take into account for later !

app/src/main/java/com/github/se/studybuddies/ui/ChatScreen.kt
cemuelle: If you mean the line 125 for the commented out part, it's just a comment to explain what the Image compose is for.
I will extract the strings in content description, but did you see elsewhere (because of your "mostly") ?

app/src/main/java/com/github/se/studybuddies/viewModels/MessageViewModel.kt
cemuelle: The default group with a list users is set because the chat is not yet linked with some real groupId that exists in Firestore, so it will crash otherwise and I decided to put a list with multiple users instead of listOf() to display a feature that have been coded.
But in the future when the chat will be correctly linked with groups and they are fully setup in the database we could remove/simplify it.

app/src/main/java/com/github/se/studybuddies/data/Message.kt
cemuelle: Yes it can be a good idea for a V2, but maybe not very in top priority.  I'm not even sure that this features has been implemented into whatsApp.

PR #92, opened 2024-04-26
Channel

hanelmouta:
I started creating the channel , channel screen

### DIFF
--- a/app/src/androidTest/java/com/github/se/studybuddies/tests/sharedtimer.kt
+++ b/app/src/androidTest/java/com/github/se/studybuddies/tests/sharedtimer.kt
@@ -1,0 +1,91 @@
+package com.github.se.studybuddies.tests
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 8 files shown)


Ovoland:
Nice start of PR but not complete enough in my opinion to be added to the main. Meaning that you should I think at least add a "route" to access this screen. Maybe I missed something but if I understood correctly there aren't any actual "route" to access this screen, which kinda lives unconnected from the rest of the app. To do so you should add a "composable" in the MainActivity and add a new Route to navigation/Route.kt (look at what have been done before for this). 
Also I have other suggestions concerning you PR. First try to create your branch from the main and not from an existing branch. You probably wrote "git switch -c "Channel" when you were on your Shared timer branch, where you should have done that in the main branch (I'm not 100% sure of that). I've done the same mistake in one of my previous PR and the issue is that you start your code on the pre-existing SharedTimerBranch. Thus if you look at the section Files changed of this PR, you can see what you implemented for the channels AND for the shared timer. In an ideal situation, you should only have your channel modification in this PR.

Ovoland (changes requested):

PR #93, opened 2024-04-26
Group invitation UI

oliver216:
This is more backend than frontend like planned.
I created a button in groupsHome to add the user to the group for which the link was generated.
For this I also created updateGroup which can be use to add any member to any group, which will be usefull for future option.


hanelmouta:
good job

hanelmouta (approved):
you did a really good job

PR #99, opened 2024-04-26
Group invitation

oliver216:
Add a link generator to create a link you can send to invite people to your group, which also shows the group name. The function to show the group name have also been created in DatabaseConnection. The link is to used on the app and not on internet like initially planned for simplification. The code to intercept an incoming link is still there but annoted for future possible use.


app/src/main/java/com/github/se/studybuddies/utility/GroupInvitation.kt
```diff
@@ -0,0 +1,64 @@
...
+suspend fun createGroupInviteLink(groupUID: String, groupName: String = ""): String {
+  return if (groupUID == "") {
+    Log.d("Link", "The Group id is empty")
+    "Group_not_founded"
+  } else {
+    var newGroupName = groupName
+    if (groupName == "") {
+      newGroupName = db.getGroupName(groupUID)
+      if (newGroupName == "") {
+        newGroupName = "NotNamedGroup"
```
Freezemoon, line 18: I feel like "Not a group name" would be better, also need to be extracted into strings.xml

Freezemoon (changes requested):
Good work ! Nice tests and such a high coverage ! I just some nitpicking (strings to extract). Also sad that the dynamic link has been given up for now but hopefully it will be picked up later on !

app/src/main/java/com/github/se/studybuddies/utility/GroupInvitation.kt
oliver216: I think I will juste put "" to avoid any confusion

Freezemoon (approved):

PR #101, opened 2024-04-26
Dark mode

oliver216:
Add the dark mode version to be always visible (so no more light mode).
Corrected a bug that crashed the app if when entering a link nothing was input. add the possibility to close this input without filling it. Change some small UI option to have a better render.


Freezemoon (commented):
Very small PR so nothing much to say, I'm just not sure why forcing the dark mode ?

oliver216:
I force the dark mode of the app (not the other interface of the phone) because we haven't coded for the light mode. (We juste have standard color for it that don't really fit and are not supposed to be used for our app.

Freezemoon:
Ohhhh i havent touched the UI so I didnt know, thanks for telling me!

Freezemoon (approved):
Looks all good!

PR #109, opened 2024-04-29
Cleaned up all hard-coded strings and put all into strings.xml

Freezemoon:


oliver216 (approved):
Very good, having all strings in the string.xml is good and will be helpful for future changes.
I checked, all needed string have been added, and added correctly.

PR #114, opened 2024-04-30
Debugged a problem when extracting strings

Freezemoon:
Just a quick PR to revert some strings that were extracted using a wrong function


Svannia (approved):
Yep, just some basic rewriting for string extraction, all good

PR #115, opened 2024-05-01
automatic APK generator

cemuelle:
automatic APK generator


Svannia (approved):
All good for me

cemuelle:
I found that it's was possible to test the APK generation without needing that it was pushed into main.
I correct the auto generation APK in this new way now !

PR #116, opened 2024-05-01
APK generator and release

cemuelle:
Automatic generate APK and create a release


cemuelle:
The end user will need to add keystore.jks and add password inside keystore.properties because we don't want to publish this files. For this reason, I also remove the file app/google-services.json of the git (we only want it on local copy and as a GitHub secret)

Ovoland (approved):
Very important modification, thank you for your great job ! It was a good idea to set up signingConfigs to get all keys and informations.

PR #129, opened 2024-05-02
Shared timer main

hanelmouta:
I have added a new timer feature to our group activities in the app. This new part helps users manage their time better when they are doing group tasks or meetings.
First, I made a new class called TimerData. This class keeps all the details about the timer like when it starts, how long it should run, if it is running right now, and how much time has passed. This is very helpful for making sure the timer works right and keeps track of time properly.
 Then made sure this timer could connect with our Firebase database. I did this by adding a function called getTimerReference in our DatabaseConnection class. This function finds the right place in the database to save or get the timer data based on the group's ID. This means that any changes to the timer get updated for everyone in the group at the same time.
Once the initial value is fetched, the countdown is managed by the device's local system. 
Another approach could involve more complex server-side logic to manage the timer, where the server would handle the countdown and only notify clients at significant events like the start, pause, or end of the timer. This could reduce client-side computation but would shift more responsibility and load to the server.But I couldn't implemented that way. 
 I implemented a method to ensure that any changes to the timer's value are detected immediately. When the timer value is first set or any subsequent change is made, it's immediately written to Firebase. Along with the timer value, the exact timestamp when this change was made is also recorded.Once the initial value and timestamp are fetched from Firebase, the application calculates the difference between the current time and the timestamp when the timer was last set. This difference represents the elapsed time since the timer was last configured.To maintain an accurate countdown, this elapsed time is subtracted from the set timer duration. For instance, if a timer was set for 60 minutes and 10 minutes have passed since it was set according to the timestamps, the remaining time calculated for the local countdown will be 50 minutes.
To make the timer readily accessible and user-friendly, I connected the timer functionality directly to the Group Home screen.
I also established a dedicated route for the timer.


app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -143,6 +145,16 @@ class MainActivity : ComponentActivity() {
...
               }
             }
+            composable(
+                route = "${Route.SHAREDTIMER}/{groupUID}",
+                arguments = listOf(navArgument("groupUID") { type = NavType.StringType })) {
+                    backStackEntry ->
+                  val groupUID = backStackEntry.arguments?.getString("groupUID")
+                  if (groupUID != null) {
+                    SharedTimerScreen(navigationActions, SharedTimerViewModel(groupUID))
+                    Log.d("MyPrint", "Successfully navigated to GroupScreen")
```
oliver216, line 155: maybe change the Lod.d message to SaredTimerScreen instead of GroupScreen

oliver216 (changes requested):
There is a bug to correct, the minus button to decremente the counter don't work.
I think the back button of the timer should go back to the group in which the timer is set, not the Home page with all group (that is a personal tought, do like you think best).

When the timer is set it's correctly put on the database, it's good. The UI is quite good, there is juste a small issus with a button depending on the phone format. No other comment on the code for that.

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
hanelmouta: done

hanelmouta: the button now can subtract

Svannia:
The timer UI should be reviewed, as Oliver said. However for me the bug is a bit different:
image

Text is not aligned, which pushed down the + button for seconds, and - button for seconds isn't displayed

oliver216:
Now the minus button works great !

I tested, when the timer is launched in a group in seams to do nothing on the other's group member's phone also connected, this could come from how you read the time from the database. (Or I misunderstood what was supposed to happen)

hanelmouta:
thanks Oliver for your review , what is suppose to happen is that each user will fetch the value of the timer and then he will be repsondable to start pause and reset the timer local

Svannia (approved):
You still have to fix that we can't "-" the seconds until its back down to 0, also we can set a negative timer.
The new UI is good

oliver216 (approved):

PR #134, opened 2024-05-03
topic setting

hanelmouta:
I implemented a settimng topic where you can delete the topic item


Svannia:
It's better with the new navigation, it works alright like this

Svannia (approved):

PR #145, opened 2024-05-03
Main correction

Ovoland:
When running the app we realized that the app was crashing when we were rotating the screen as it was creating another StreamVideo instance. Thus we forced the app to be only on portait mode (as it was also decided few weeks ago). Plus I also solved some issue with the LauncedEffect. I also fixed the issues with the new folder organization. And finally I reformated the code as the previous PR were unfortunately accepted while the CI wasn't passing....
I didn't have the time but we have a big security issue with the API key on the main. This should be fixed later


Svannia:
Okay all good

PR #150, opened 2024-05-03
Small fixes

Ovoland:
I corrected small issues remaining in the main. For example, there were some issues with the map which wasn't working properly on Android 14 (my tablet was an Android 13 so I didn't notice this earlier). Also we prevent the user from clicking again on a bottom navigation icon if he's already on the icon page (prevent from loading again the viewmodel and everything). Also I blocked the path to the EditToDo as this screen wasn't really working.


Svannia (approved):
Great that you found the fix for the map on newer API versions! Also good with the bottom navigation disable

PR #153, opened 2024-05-06
Location Database and friend location

Ovoland:
I allow the user location to be added to the database when the location tracking is on. This means that the location is set "offline" when the location is off or when the user is closing the app (even after having tracked his location). The default location is "offline" when a user is created. 
I also have added the user friend's location to the map. The user can in any time (even if it's own location isn't tracked) see where are his friends (if they have set the location tracking on on their app). 
Also I remind that the location tracking is active in the background of the app. Thus the user don't have always be on the map screen to have his location tracked.


app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -213,6 +229,16 @@ class DatabaseConnection {
         }
   }
 
+  fun updateLocation(uid: String, location: String) {
```
Svannia, line 232: There are a few places (like here) when you need the location as a string. Given that the two possible states of the location are "online" or "offline", maybe create an enum class LocationState or sth and have it contain OFFLINE and ONLINE. This way, when you need to make checks, you can just have a switch statement over the types of LocationState and there can't be any bug because of mistyping

app/src/main/java/com/github/se/studybuddies/ui/map/MapScreen.kt
```diff
@@ -123,46 +152,140 @@ fun MapScreen(
                       snippet = "This is your current location")
                 }
               }
+              friendsData.value.forEach { friend ->
+                if (friend.location != "offline" && friend.location != "") {
```
Svannia, line 156: For example here, instead of writing offline as a hard-coded string yourself, you'd just check `friend.location != LocationState.OFFLINE`

If you think it's a good idea, you can just put the enum in User.kt

Svannia (changes requested):
Good job with the DB and (from what I can tell so far) the permissions 👍

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
Ovoland: Thanks for your comment but maybe the code wasn't clear enough or you misunderstood something. But there aren't just 2 states "offline" and "online" for the location. What I thought was to set the location to "offline" when the user isn't tracking its location and he closes the app. On the other hand, we the user is using the app, the location it set to its current location. I give this location as string in updateLocation because it was easier. Thus the location should either be "offline" or a location (for example "40.0034,78.02323"). I hope this was clear. Do you have any other suggestions ?

Svannia: oooh okay I understand! In that case maybe just extract the string like the others to avoid hard-coding it and all good !

Svannia (approved):
all good !

PR #158, opened 2024-05-07
save button correctly has disabled colour

Svannia:
- removed duplicate codes to create save buttons
- group and topic creation have correct save button with disabled colour
- group and topic creation background fix

- quick project hierarchy re-ordering, those directories were going places XD (not in a good way)
- put same save button as everywhere in group settings
- fixed db mistake in group settings I saw while doing the save button
- minor UI fix in settings pages


Freezemoon (approved):
Seems good to me ! Very good global reorganisation of the overall project. The code is simple and efficient. I just noticed some unused imports and variables that were issued by Sonar but otherwise all the rest is good !

PR #159, opened 2024-05-08
Topic chat

cemuelle:
Implement Topic Chat, small optimizing in DirectMessageScreen
And adding tests


oliver216 (approved):
I tested the functionality, it works great on my machine. The code is cleaner than before, using less "if" which decreased complexity and give a better view of what is happening. The new code is good, I have nothing to comment on that.

The test presents are good, but don't cover enough code. Think about adding more test.

PR #160, opened 2024-05-08
Timer backgroup

hanelmouta:
I fixed the timer.First of all , I changed the hall implementation of the timer. I used the countdown class. In order to make it run in the background , I found the idea to make the timerviewmodel a singleton class so only one instance is created. Before each time the screen was accessed it created a new view model


app/google-services-3.json
cemuelle, line 1: You don't want to commit/push this file. It should stay secret (on local + setup as secret in CI)

hanelmouta: thanks for pointing that out I delete the copy

Freezemoon (changes requested):
Hi ! I tested the build and I noticed two small problems : 
1. When hitting pause, the timer was effectively paused, but it's not clear how to unpause it ? I tried clicking on pause again, it does nothing, and clicking on start would restart the time from the latest input time. 
2.  The solo timer and the group timers seem to be linked ? I started a solo timer in the background and wanted to test the group timer but it was actually the solo timer just counting down and hitting the go back button led me to the solo study home again. From my understanding, the group timers should be separate from the solo timers.

app/src/main/java/com/github/se/studybuddies/ui/timer/TimerScreenContent.kt
```diff
@@ -32,21 +32,25 @@ import com.github.se.studybuddies.ui.theme.Blue
...
-      navigationActions = navigationActions,
-      timerValue = timerValue,
-      timerEnd = timerEnd,
-      onAddHours = { hours: Long -> timerViewModel.addHours(hours) },
-      onAddMinutes = { minutes: Long -> timerViewModel.addMinutes(minutes) },
-      onAddSeconds = timerViewModel::addSeconds,
-      onStart = { timerViewModel.startTimer() },
-      onPause = { timerViewModel.pauseTimer() },
-      onReset = { timerViewModel.resetTimer() })
+  timerValue?.let {
```
Freezemoon, line 40: Can delete those two null-safe accesses as they won't ever be null, above and below this comment

app/src/main/java/com/github/se/studybuddies/viewModels/TimerViewModel.kt
```diff
@@ -1,61 +1,90 @@
...
-  fun addSeconds(seconds: Long) {
-    if (_timer.value + seconds > 0) {
-      _timer.value += seconds
-    }
+  private val _timerEnd = MutableStateFlow(false)
+  val timerEnd: StateFlow<Boolean> = _timerEnd
+
+  private var countDownTimer: CountDownTimer? = null
+
+  private fun setupTimer(duration: StateFlow<Long>) {
```
Freezemoon, line 31: this duration parameter is never used

app/src/main/java/com/github/se/studybuddies/ui/timer/TimerScreenContent.kt
hanelmouta: it is for safety reasons they won't let me delete it

app/src/main/java/com/github/se/studybuddies/viewModels/TimerViewModel.kt
hanelmouta: I delete it

hanelmouta:
Thank you for your review , I solved the issue with the pause. For the second part of your review there is a pr that solve the issue of the shared timer

Freezemoon (approved):

PR #161, opened 2024-05-08
Chat functions

cemuelle:
Add possiblity to send extra type of data as Image in chats

Modify/move shared UI fonctions to be more general and be able to use them in my case.


hanelmouta (approved):
Hi , you did an amazing job , the good look's fine for me I tried adding an image and it's working 

Screenshot 2024-05-09 at 23 42 28

PR #162, opened 2024-05-08
Change group name

Svannia:
just a small mistake from my part that made bugged group re-naming.
A real quick fix tho


oliver216 (approved):
Seems like a very good fix indeed, the bug was quite annoying, great job !

PR #163, opened 2024-05-09
Topic items display

Svannia:
- differentiate the display of topic items for folders and files
- can add folders/files inside an existing folder
- sorting lists by file first, then folders, and each sublist by descending alphabetical order
- considered end cases to avoid creating an item in the wrong folder or having multiple menus/buttons piling up on the screen
- prepared the DB to add the deletion feature for topic items


cemuelle (changes requested):
Great code ! If you extract the const in database and stop using material and material3 at the same time in TopicScreen it will be nice.

However during testing of the app, I found some behaviours that are special.
- When we create a subfolder, when expanded we cannot make the difference if it's a folder or a subfolder
- When we add a file into a folder, it is added into the folder as we expect, but also on the top list with all the files that are not contained into any folders.
- When we add a file into a folder, the folder close itself and to see however newly created file we need to re expand the folder
- When we want click on add folder or file, the input text is very similar in term of colors of the list of folders/files and just appear on top. I suggest to use an AlertDialog, this will also allow to easly add a save button and allow dissmiss if we don't want anymore to create a new item.
- Last point, when we only have files on top that don't belongs to a folder, should we still need the indentation ?

Except theses behaviours that are unfamiliar for me, it's working well !
Good job!

Svannia:
@cemuelle thanks a lot for testing the feature like that, your comments actually gave me ideas for display problems (such as subfolders) that I didn't know how to tackle.
I'm gonna make, in this PR, the changes that are related to good functionality. For user experience upgrade modifications, I'll do them in another PR

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -708,13 +708,13 @@ class DatabaseConnection {
...
 
-  // using the topicData collection
+  // using the topicData and topicItemData collections
   suspend fun getTopic(uid: String): Topic {
     val document = topicDataCollection.document(uid).get().await()
     return if (document.exists()) {
-      val name = document.getString("name") ?: ""
-      val exercisesList = document.get("exercises") as List<String>
-      val theoryList = document.get("theory") as List<String>
+      val name = document.getString(Companion.topic_name) ?: ""
```
cemuelle: Nice modification !
To nitpick, could you either put everywhere Companion.xxx or remove the Companion here ?

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -739,19 +739,17 @@ class DatabaseConnection {
     for (itemUID in uids) {
       val document = topicItemCollection.document(itemUID).get().await()
       if (document.exists()) {
-        val name = document.getString("name") ?: ""
-        val parentUID = document.getString("parent") ?: ""
-        val type = ItemType.valueOf(document.getString("type") ?: ItemType.FILE.toString())
+        val name = document.getString(Companion.topic_name) ?: ""
```
cemuelle: Same for this

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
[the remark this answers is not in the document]
cemuelle: Nice job to put the string into const !

cemuelle (approved):

PR #164, opened 2024-05-09
Reformatted UI of Create Account and solved a bug 

Freezemoon:
- Reformat CreateAccount UI
- Fixed a bug where when closing app during account creation and coming back would cause the app to crash
- Fixed a follow-up bug where when coming back, would allow into SoloStudy with empty username and profile picture
- Now, coming back without saving a username will navigate to CreateAccount


Svannia (approved):
All good with me, the screen looks good. The "empty account" bug when closing the app after loging in but before creating an account is nicely fixed, no more empty usernames in the DB haha. Good job !

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -99,7 +99,17 @@ class MainActivity : ComponentActivity() {
...
+                db.userExists(
+                    uid = db.getCurrentUserUID(),
+                    onSuccess = { userExists ->
+                      if (userExists) {
+                        navController.navigate(Route.SOLOSTUDYHOME)
+                      } else {
+                        navController.navigate(Route.CREATEACCOUNT)
+                      }
+                    },
+                    onFailure = { Log.d("MyPrint", "Failed to check user existence") })
```
cemuelle, line 112: Nice check, however what happen if the the call to the database is a failure, we got a Log but we block the user.

It's a big issue because we are making the offline functionality unusable without internet. The user is blocked into a blue screen !

Additionally, it will be cleaner if you move your check into UserViewModel and then call the viewModel.

app/src/main/java/com/github/se/studybuddies/ui/account/CreateAccount.kt
```diff
@@ -71,30 +82,56 @@ fun CreateAccount(userViewModel: UserViewModel, navigationActions: NavigationAct
...
+                          }
+                          .testTag("go_back_button"))
+            },
+            actions = {})
+      }) { paddingValue ->
+        Column(
+            modifier = Modifier.fillMaxSize(),
+            horizontalAlignment = Alignment.Start,
+            verticalArrangement = Arrangement.Top) {
+              LazyColumn(
```
cemuelle, line 107: Did we really need to be able to scroll ?
If not, we should remove the Lazycolumn and directly display our elements

app/src/main/java/com/github/se/studybuddies/ui/account/LoginScreen.kt
cemuelle, line 1: Nice to extract strings

cemuelle (changes requested):
Glad that you optimize the code as extracting some strings, remove useless call and fix crashes.
However with you're modification it's blocking the access the offline features when we don't have any internet connection ! You need to correct that before I can accept this PR

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -90,7 +90,7 @@ class MainActivity : ComponentActivity() {
                       context = context,
                       apiKey = apiKey, // demo API key
                       geo = GEO.GlobalEdgeNetwork,
-                      user = User(id = db.getCurrentUser().username),
+                      user = User(id = "Test"),
```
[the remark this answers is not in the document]
Freezemoon: I'm just putting this here to avoid an error, I actually changed all of this in another one of my branches so those lines won't be here anymore

app/src/main/java/com/github/se/studybuddies/ui/account/CreateAccount.kt
Freezemoon: I followed the AccountSetting formatting to uniformize everything

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
Freezemoon: I solved the offline issue, it should be good now. However, for the viewModel it just seems very complicated for such a simple thing, especially since the database is already instanciated in the main and I'm just calling its functions.

PR #165, opened 2024-05-09
Shared timer correction

hanelmouta:
new version of the shared timer that now runs in the background


app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -264,7 +264,9 @@ class MainActivity : ComponentActivity() {
                     backStackEntry ->
                   val groupUID = backStackEntry.arguments?.getString("groupUID")
                   if (groupUID != null) {
-                    SharedTimerScreen(navigationActions, SharedTimerViewModel(groupUID))
+                    val viewModel2 = SharedTimerViewModel.getInstance(groupUID)
```
Ovoland, line 267: Maybe you could write SharedTimerViewModel.getInstance(groupUID) inside of the SharedTimerScreen ? I don't think it's so important but it saved from the creation of variable that is really used elsewhere

app/src/main/java/com/github/se/studybuddies/viewModels/SharedTimerViewModel.kt
```diff
@@ -1,203 +1,332 @@
...
-  private fun listenToTimerUpdates() {
-    timerEventListener =
-        object : ValueEventListener {
-          override fun onDataChange(snapshot: DataSnapshot) {
-            val timerValue = snapshot.getValue(Long::class.java) ?: 0L
-            remainingTime.postValue(timerValue)
-            timerData.postValue(TimerData(System.currentTimeMillis(), timerValue, 0L))
-            synchro()
-            pauseTimer()
+  private fun subscribeToTimerUpdates() {
```
Ovoland, line 35: You're not using this function in your code and you could probably remove it

app/src/main/java/com/github/se/studybuddies/viewModels/SharedTimerViewModel.kt
```diff
@@ -1,203 +1,332 @@
...
+      groupUID?.let { uid ->
+        val newEndTime = System.currentTimeMillis() + newTime
+        viewModelScope.launch { databaseConnection.updateGroupTimer(uid, newEndTime, isRunning) }
+      }
+    }
+  }
+}
+
   /*
+    private val databaseConnection = DatabaseConnection()
```
Ovoland, line 143: You have a lot of functions that are commented at the end of your file. If you don't need them anymore you should also probably removed them in this PR

Ovoland (changes requested):
Amazing job :D I'm really impressed by the way you handle the timer in background ! You used another technic that what I did for the map and it's super interesting to discover yours which seems shorter and easier. The code is short and super efficient, and I haven't spotted any bugs. The timer seems to be handled correctly when we leave a group do other things in the app and come back again. Same for closing the phone. -> Background property seems to work perfectly. I've marked few recommendations in the PR, but it really just small fixes to improve few points. Before merging this PR, I would just suggest you to remove all the unnecessary part of your code (the commented lines) and check my comments. Also I have a last suggestion concerning the UI. The timer test doesn't exactly have the app color theme. It's a bit tricky but to have the correct color you should write color = Blue, instead of color = Color.Blue. The difference lies in the fact that Color.Blue is a predefined color of Android studio, whereas Blue is a color that we have personally defined for the project.  You can check the photo to see where you have to change the code (It's lines 62 of SharedTimerScreen): 
image
image
I don't have any further remarks on the current code which is in my opinion really good and seems to work, but I have a suggestion for a later feature. It would be nice to have something to inform the user that the timer is done, even if the user is in another screen (maybe using notification or phone vibration).
That's all I have to say. Well done :)

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
hanelmouta: I prefer having a variable because if in thee future I make changes it is better to change just the variable than all places where it is used

Ovoland (approved):

PR #170, opened 2024-05-09
Leave group

oliver216:
Add a function to remove a user from a group, which will also be deleted if there is no user left in the group.

Add an option to leave a group. (UI)

Add an confirmation panel to be sure if you want to leave the group or cancel.

add test for this function.

Small other cleaning and bug fix.


TdVBd (approved):
Code is clean, leave group functionality woeking as expected

Maybe the "Yes" button could be red with white text instead, as the blue-red contrast looks a bit harsh to the eye

PR #173, opened 2024-05-10
Todolist bug fix

TdVBd:
solved a bug causing a crash when trying to edit a task and moved the encryption scheme methods into a separate class


hanelmouta (approved):
Hey , good job on fixing the bug, I was able to create a task and modify it.
Screenshot 2024-05-10 at 18 19 24
Good job on working on the encryption.However the refactoring around encryption, including the handling of secret keys in SharedPreferences, raises concerns about security best practices. Storing encryption keys on the device, even if base64 encoded, could pose security risks, you should keep working on that. 
Overall you did a good job

PR #175, opened 2024-05-10
Placeholder screen

TdVBd:


oliver216 (approved):
There is no description of your pull request, it would be useful for me to better understand your change.

The code is clean and understandable, I don't see any bad practice to change. Also, there are a lot of comments which are good to understand the code, great.

This PR is less big than the last you did which is also good to review it.

The application works well on my machine, when testing on my side I saw a few UI bugs but I'm not sure they are all linked directly to this PR : 
When you have finished writing in the description section, the keyboard doesn't disappear which hide the save button.
The blue color you use if from Color.Blue and not just Blue like it should to match the color pattern of our app.
The date doesn't seam to be saved correctly.
The rest is clean, well made, and intuitive, good work.
(I approve the PR anyway because not sure this need to be addressed here)

There is no new issuses or anything from sonar which is good, but the coverage is too low.

PR #176, opened 2024-05-11
Contact list

TdVBd:
Implemented the utilitary database functions for the Contact List feature aswell as the corresponding data class and viewmodel
Modified the DirectMessageScreen, now when starting a new chat with a user from the userlist, a new contact is created between it and the current user


app/src/main/java/com/github/se/studybuddies/data/Contact.kt
```diff
@@ -0,0 +1,24 @@
...
+data class Contact(val id: String, val members: Pair<String, String>, val showOnMap: Boolean) {
+  companion object {
+    fun empty(): Contact {
+      return Contact(id = "", members = Pair("", ""), false)
+    }
+  }
+}
+
+class ContactList(private val contacts: List<Contact>) {
+  fun getAllTasks(): List<Contact> {
```
hanelmouta, line 12: getAllTasks should be renamed to getAllContacts for clarity.

hanelmouta, line 12: why do you have this companion function ?

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -1112,6 +1125,101 @@ class DatabaseConnection {
...
+      val members = document.get("members") as? Pair<String, String> ?: Pair("", "")
+      val showOnMap = document.get("showOnMap") as Boolean
+      Contact(contactUID, members, showOnMap)
+    } else {
+      Log.d("MyPrint", "contact document not found for contact id $contactUID")
+      Contact.empty()
+    }
+  }
+
+  suspend fun createContact(otherUID: String) {
```
hanelmouta, line 1166: I think you can create two time the same contact is there a way to ensure that only one contact of the same person is created ?

app/src/main/java/com/github/se/studybuddies/viewModels/ContactViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.github.se.studybuddies.data.Contact
+import com.github.se.studybuddies.database.DatabaseConnection
+import kotlinx.coroutines.launch
+
+class ContactViewModel(private val contactID: String? = null) : ViewModel() {
```
hanelmouta, line 11: Can't you regroup the two viewModels?

hanelmouta (approved):
Hi , I reviewed your pr , you did a really good job. Your code is clean and well written . They are no major change in the ui but I check that everything is visible in the firebase and it is.  There is a small issues and it is that you can create a contact twice of the same person.

app/src/main/java/com/github/se/studybuddies/data/Contact.kt
hanelmouta, line 12: The empty() method in the companion object of Contact might be redundant if not used.

TdVBd: you can check its usages

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
TdVBd: this bug has now been solved

app/src/main/java/com/github/se/studybuddies/viewModels/ContactViewModel.kt
TdVBd: yes i might do that

PR #177, opened 2024-05-11
Map bug fix

Ovoland:
During last friday demo, I realized that there was a bug with the map. It was related to the database and the way the location string was sent. I found that it was due the use of "R.string.offline". Thus I went back to my first idea of hardcoding  "offline" and it solved the issue


Freezemoon (approved):
Simple bugfix that was solved the easiest way, good job !

PR #178, opened 2024-05-12
Solo calendar 

hanelmouta:
I implemented the calendar screen

### DIFF
--- a/app/src/androidTest/java/com/github/se/studybuddies/tests/CalendarAppTest.kt
+++ b/app/src/androidTest/java/com/github/se/studybuddies/tests/CalendarAppTest.kt
@@ -1,0 +1,58 @@
+package com.github.se.studybuddies.tests
+
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.se.studybuddies.calender.CalendarApp
+import com.github.se.studybuddies.navigation.NavigationActions
+import com.github.se.studybuddies.screens.CalendarScreen
+import com.github.se.studybuddies.viewModels.CalendarViewModel
+import com.kaspersky.components.composesupport.config.withComposeSupport
+import com.kaspersky.kaspresso.kaspresso.Kaspresso
+import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
+import io.github.kakaocup.compose.node.element.ComposeScreen.Companion.onComposeScreen
+import io.mockk.impl.annotations.RelaxedMockK
+import io.mockk.junit4.MockKRule
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class CalendarAppTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withComposeSupport()) {
+
+  @get:Rule val composeTestRule = createComposeRule()
+
+  @get:Rule val mockkRule = MockKRule(this)
+
+  @RelaxedMockK lateinit var mockNavActions: NavigationActions
+
+  @Before
+  fun testSetup() {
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/github/se/studybuddies/calender/CalendarScreen.kt
```diff
@@ -0,0 +1,189 @@
...
+            },
+            actions = {})
+      },
+  ) { padding ->
+    Surface(
+        modifier = Modifier.fillMaxSize().verticalScroll(rememberScrollState()).padding(padding)) {
+          CalendarWidget(
+              days = DateUtil.daysOfWeek,
+              yearMonth = uiState.yearMonth,
+              dates = uiState.dates,
```
Ovoland, line 68: Instead of passing yearMonth and dates, you could just give uiState as parameter and take extract yearsMonth and dates inside of the CalendarWidget

app/src/main/java/com/github/se/studybuddies/calender/CalendarScreen.kt
```diff
@@ -0,0 +1,189 @@
...
+              onPreviousMonthButtonClicked = { prevMonth -> viewModel.toPreviousMonth(prevMonth) },
+              onNextMonthButtonClicked = { nextMonth -> viewModel.toNextMonth(nextMonth) },
+              onDateClickListener = {
```
Ovoland, line 71: Ok I think a bit and I think they could probably be a better way to initialize CalendarWidget function. The parameters yearsMonth,dates, onPreviousMonthButtonClicked, onNextButtonClicked (and maybe onDateClickListener), they all need/are initialize by the viewModel. Then instead of passing all these different parameters inside of the CalendarWidget, I will just pass the viewModel as a parameter. Then you could initialize the other parameters later. 
Also I realized that you aren't using yearMonth, dates,onPreviousMonthButtonClicked and onNextMonthButtonClicked on the CalendarWidget. You're only taking these parameters to give them to Header and Content. Thus I will as well just give the viewModel to Header and Contente instead of all the other parameters. Therere yearMonth, dates,onPreviousMonthButtonClicked and onNextMonthButtonClicked will only be created when needed in the appropriate function at the right timing. 
What do you think of it ?

app/src/androidTest/java/com/github/se/studybuddies/tests/CalendarAppTest.kt
Ovoland, line 1: You could probably also make tests of the top app bar if you have time

Ovoland (commented):
Well done, this first draft of the calendar seems to work well :D Plus you also have included tests which is nice. Check my comments on the code for more details :)

Ovoland (commented):
reviewed

app/src/main/java/com/github/se/studybuddies/calender/CalendarScreen.kt
hanelmouta: Hi, thanks you for your comment.  it is just a way of coding , instead of instantiating the variable in the CalendarWidget I do it in CalendarApp so that if I want to reuse calendar Widget for other function I can, for exemple if I code the shared calendar I can reuses this function in the screen of the shared variable , it is just for modularity because I reuse a lot of the function in other classes , and I am passing the parameter of the view model prevMonth -> viewModel.toPreviousMonth(prevMonth)

app/src/androidTest/java/com/github/se/studybuddies/tests/CalendarAppTest.kt
hanelmouta: I will do it later in another pr if I have time

app/src/main/java/com/github/se/studybuddies/calender/CalendarScreen.kt
hanelmouta: hey it doesn't change anything , I prefer being explicit on the parameter I need for the function

PR #180, opened 2024-05-13
Delete group

oliver216:
Add a function deleteGroup that will delete the group from all users in it, delete the picture in storage and delete the group.
(To be added later, also delete chat and topic)

Add the UI to delete the group

Add UI Display to ask confirmation if user want really want to delete the group and a warning that it deletes it for all users.

Change small UI, like the button to confirm to leave the group become red instead of the text that was red.


Ovoland (approved):
Good job ! The group deletion and leaving seem to work well ! Plus it's nice that you thought of deleting the group if there are no more members after someone left.  
The code seems correct to me and I don't really have any other suggestions concerning it :)
However I have some other suggestions for later features and small adjustments. For example, after leaving or deleting a group, the screen is refreshing and for few seconds we can see the message when we don't have group (our group list is loading). Thus it would be nice to add a waiting circle for few seconds after interacting with the group. 
Also It would be nice if you could the same "modify, leave, delete" option whens we're on the group page. Check the picture below: 
groupSwent
That's all
Good job :D

PR #184, opened 2024-05-13
Improve chat

cemuelle:
- Correct error in topics chat that messages weren't displayed.
- Correct bug when rotating screen
- Correct UI bugs that list of direct messages wasn't able to take full size.
- Optimize list of friends and direct messages in order to reduce/supress loading time and not reload at each time, and also directly display when someone start a new chat without reloading
- Correct bug with message blinking and moving left/right on loading 
- Optimize getMessage in database and resolve sonar issue

Also fix blinking of other parts of the app (that are not related to the chat, but the fix was the same)


cemuelle:
Correct the issue were the messages were blinking at the loading and moving left/right

cemuelle:
I profit to correct the blinking of other parts of the app that were also blinking in this PR (like in groups) because it's the same fix that for the message and it take only few lines.

Freezemoon (approved):
Looks pretty neat to me ! Good job on solving the bug of the "blinking" screens as it was a reoccuring instance and was somewhat annoying. I really like the way you manage the different types of data that a user can send (file, picture, simple msg etc). The code is very clean and easy to understand. The only ick is that the coverage is pretty low but since its mostly modifying old code, it's not that important since the CI visibly passed. Nice job !

PR #185, opened 2024-05-15
Group UI update

oliver216:
Add the hamburger menu to modifyGroup and to the screen when in a group

Correct small bug like "enter link" text zone being too big when no group existed

Small UI Improvement (smoother button and so on)

Add to userRemoveGroup when deleting an empty group deletion of chats and pictures

Add to deleteGroup the deletion of the chats

Reduce the complexity of the MainActivity


TdVBd (approved):
Great job overall, all the added corrections are behaving as expected
One remark, the hamburger menu when inside the modifyGroup screen has "modify group" amongst its options, which simply leads back to the same screen and seems unecessary

PR #186, opened 2024-05-15
Topic improvement

Svannia:
Following the reviews from @cemuelle 
- Adding a new item expands its parent folder
- Indentation is based on hierarchy depth
- Field to input new item name is an AlertDialog

### DIFF
--- a/app/src/main/java/com/github/se/studybuddies/ui/topics/TopicScreen.kt
+++ b/app/src/main/java/com/github/se/studybuddies/ui/topics/TopicScreen.kt
@@ -25,6 +25,7 @@ import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Add
 import androidx.compose.material.icons.filled.MoreVert
+import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.Button
 import androidx.compose.material3.DropdownMenu
 import androidx.compose.material3.DropdownMenuItem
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/se/studybuddies/ui/topics/TopicScreen.kt
```diff
@@ -406,12 +434,21 @@ fun AddInFolderButton(
...
-    parentUID: MutableState<String>
+    childrenExpanded: MutableState<Boolean>,
+    parentUID: MutableState<String>,
+    depth: Int
 ) {
   val expandedState = remember { mutableStateOf(false) }
-  Box(modifier = Modifier.fillMaxWidth()) {
+  val screenWidth = LocalConfiguration.current.screenWidthDp
+  val offset =
+      if (depth > 0) DpOffset((screenWidth - 180).dp, 0.dp) else DpOffset(x = (-16).dp, y = 0.dp)
```
cemuelle, line 444: In the first DpOffset, you didn't give the parameter value is assigned to, but in the second yes (x = ..., y= ...). Try to stay consistent between the two (either but for both, or for none)

Additionnaly you could make the line more readable by not calling DpOffset twice for only one difference.
You could do this instead : `val offset = DpOffset((if (depth > 0) (screenWidth - 180) else (-16)).dp, 0.dp)` but it still no very easy to read, so I recommend to extract it into a new value :
```
val xOffset = if (depth > 0) (screenWidth - 180) else (-16)
val offset = DpOffset(xOffset.dp, 0.dp)
```

cemuelle (approved):
Great job,
This correct all the previous issues that were preventing to fully enjoy create/navigating into the topics. 
The use of an alertDialog help to see that where we should write the name and also allow to cancel in case we missclick before ! Just one point the padding is a little to big around the textfield, it makes also the buttons to far from the textfield.
And you code is clean and easy to read. I just suggest to modify the offset as commented.
Last point, the coverage is inexistant, if you can and some tests it will be perfect.

PR #189, opened 2024-05-16
Calendar Event 

hanelmouta:
event display


app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -137,8 +139,13 @@ class MainActivity : ComponentActivity() {
...
-                val calendarViewModel = remember { CalendarViewModel() }
-                CalendarApp(calendarViewModel, navigationActions)
+                val calendarViewModel = remember {
+                  auth.currentUser?.let { it1 -> CalendarViewModel(it1.uid) }
+                }
+                if (calendarViewModel != null) {
+                  CalendarApp(calendarViewModel, navigationActions)
+                }
+
                 Log.d("MyPrint", "Successfully navigated to GroupsHome")
```
Freezemoon, line 149: The log hasn't been adapted to show that it navigated to the Calendar

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -163,6 +170,17 @@ class MainActivity : ComponentActivity() {
...
                 }
+            composable(
+                route = "${Route.DAILYPLANNER}/{date}",
+                arguments = listOf(navArgument("date") { type = NavType.StringType })) {
+                    backStackEntry ->
+                  val date = backStackEntry.arguments?.getString("date")
+                  val currentUser = auth.currentUser
+                  if (date != null && currentUser != null) {
+                    DailyPlannerScreen(date, CalendarViewModel(currentUser.uid), navigationActions)
+                    Log.d("MyPrint", "Successfully navigated to Settings")
```
Freezemoon, line 181: This one as well hasn't been adapted to correctly report where we are navigating

app/src/main/java/com/github/se/studybuddies/viewModels/CalendarViewModel.kt
```diff
@@ -1,23 +1,106 @@
...
+  private val databaseConnection = DatabaseConnection()
+
+  private val _dailyPlanners = mutableStateMapOf<String, MutableStateFlow<DailyPlanner>>()
+  val dailyPlanners: Map<String, StateFlow<DailyPlanner>>
+    get() = _dailyPlanners
+
+  private val _uiState = MutableStateFlow(CalendarUiState.Init)
+  val uiState: StateFlow<CalendarUiState> = _uiState.asStateFlow()
+
+  init {
```
Freezemoon, line 29: From what the coach told us on the general chat on discord, using init in a viewModel is a bad practice and should be avoided. I'll let you have a look at our conversation on discord.

app/src/main/java/com/github/se/studybuddies/viewModels/CalendarViewModel.kt
```diff
@@ -1,23 +1,106 @@
...
+    viewModelScope.launch {
+      _uiState.update { currentState ->
+        currentState.copy(yearMonth = prevMonth, dates = dataSource.getDates(prevMonth))
+      }
+    }
+  }
+}
+
+/*
 class CalendarViewModel : ViewModel() {
```
Freezemoon, line 84: Huge chunk of code that should be deleted if unused

Freezemoon (changes requested):
Good lengthy code overall, pretty long PR tho. The display of the daily planner is nice but kinda counter-intuitive. There are lines that we can click and it makes the keyboard pop but can't write anything to it and to add items, we have to click on the + sign on the side. I feel like it would be easier to either make the lines unclickable or get rid of them totally and only put the + sign. The save button doesn't work after filling in the page and the first time I tested it,  I wrote "invisible" items by clicking on the lines and once all three fields had been "filled", clicking save would crash the app. The display of the calendar is very nice and works well ! Good idea to also display the current day on the calendar ! Additionally, some tests would be welcome as they aren't any. Otherwise, good job as it's a somewhat complicated feature to implement in a week.

app/src/main/java/com/github/se/studybuddies/viewModels/CalendarViewModel.kt
hanelmouta: will change it in another pr

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
```diff
@@ -139,9 +141,14 @@ class MainActivity : ComponentActivity() {
...
               ifNotNull(remember { auth.currentUser }) { _ ->
-                val calendarViewModel = remember { CalendarViewModel() }
-                CalendarApp(calendarViewModel, navigationActions)
-                Log.d("MyPrint", "Successfully navigated to GroupsHome")
+                val calendarViewModel = remember {
+                  auth.currentUser?.let { it1 -> CalendarViewModel(it1.uid) }
+                }
+                if (calendarViewModel != null) {
+                  CalendarApp(calendarViewModel, navigationActions)
+                }
```
RaulinN, line 149: Careful you introduced a big jetpack compose mistake in these few lines! Can you spot it?

<details>
  <summary>Spoiler warning</summary>
  
  How do you create [view models with dependencies in Jetpack compose](https://developer.android.com/topic/libraries/architecture/viewmodel/viewmodel-factories#jetpack-compose)?
  
</details>

RaulinN (commented):
Did not review the full PR in case some of you still need PR reviews for M3

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
hanelmouta: I used a factory thank you for your comment and the useful link

RaulinN:
Btw small thing I noticed @hanelmouta; you are often doing a lot of short commits (sometimes removing a single space or line break, e.g. [136fdce](https://github.com/Study-Buddies-SwEnt/SwEnt_Group_project/pull/189/commits/136fdce525117140754c4bb47c71ad2175579794)).

If you *already* committed something (but did not push it yet), you can **amend** the last commit you did, i.e. add or remove stuff from the last un-pushed commit. For instance,

```bash
# add all the modified files (file.txt) in a commit w/ message "commit message 1"
git add file.txt
git commit -m"commit message 1"
# you then realise you forgot to add a file in your commit, you can add it by using the following commands
git add correction.txt
git commit --amend
# by doing this, you can rename the first commit, and your git history will show a single commit instead of 2
```

Freezemoon (approved):

Freezemoon (changes requested):
Correct what the coach pointed out higher up

Freezemoon (approved):

PR #190, opened 2024-05-16
Add file and link in Chat

cemuelle:
This PR increase the possibility of messages of the chat :

- Link messages and when the user click on them he's redirected to his browser and the webpage
- File messages, the user can select a file present on his phone, send it. And then when we want to open it, he can choose which app to open the pdf.


app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
```diff
@@ -504,3 +597,115 @@ fun SendPhotoMessage(messageViewModel: MessageViewModel, showAddImage: MutableSt
...
+
+  val getContent =
+      rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) { uri: Uri? ->
+        uri?.let { fileUri ->
+          fileState.value = fileUri
+          val cursor = context.contentResolver.query(fileUri, null, null, null, null)
+          cursor?.use {
+            if (it.moveToFirst()) {
+              val nameIndex = it.getColumnIndex(OpenableColumns.DISPLAY_NAME)
+              if (nameIndex != -1) {
```
oliver216, line 666: You could remove one of the nested loop here, by directly affecting nameIndex in the if statement and checking != -1

app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
```diff
@@ -288,19 +347,24 @@ fun OptionsDialog(
...
+                        showOptionsDialog.value = false
+                      }) {
+                        Text(
+                            text = stringResource(R.string.edit),
+                            style = TextStyle(color = White),
+                        )
+                      }
+                  Spacer(modifier = Modifier.height(8.dp))
+                }
+                else -> {}
```
oliver216, line 367: Should this really be empty ?

oliver216 (approved):
Good PR, the code is clean and understandable, you use a lot of when case instead of if which is good for complexity and comprehension. The PR is big, try not to do bigger than this.

I tested the functionality, work great and are intuitive, good job !

There is only some small improvement you can do, I noted it directly on the concerned place. Could be great to do more tests for this part.

app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
```diff
@@ -504,3 +597,115 @@ fun SendPhotoMessage(messageViewModel: MessageViewModel, showAddImage: MutableSt
...
   }
 }
+
+@Composable
+fun SendLinkMessage(messageViewModel: MessageViewModel, showAddLink: MutableState<Boolean>) {
+  val linkState = remember { mutableStateOf("") }
+  val linkName = remember { mutableStateOf("") }
+
+  if (showAddLink.value) {
+    AlertDialog(
```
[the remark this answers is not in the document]
cemuelle: What I understand between AlertDialog and Dialog, it's that AlertDialog are easier to implement, but doesn't allow for more advance customizations.
So I choose the simplier way because I don't need the extra customization

app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
cemuelle: If you speak about the else -> { }, yes because I'm using a when, but not all case are defined. And I want to do nothing with the case I don't defined clearly something to do.

oliver216 (approved):
Great modification ! It's a lot cleaner now.

PR #194, opened 2024-05-16
Improvement of the End to End test

Ovoland:
In order to implement End to End tests and to facilitate the tests in general, we decided to integrate a Fake database in our project. That's what I was working on.


app/src/main/java/com/github/se/studybuddies/database/DbRepository.kt
RaulinN, line 1: Yes great! For future projects (this would probably take too much refactoring here), I'd suggest splitting this file in multiple repositories interfaces. For instance `IUserRepository`, `IGroupRepository`, `ITopicRepository`, ... This way your interface does not have 50.000 functions and you can have better compartmentalization

For instance,

```kotlin
class GroupsViewModel(private val db: IGroupRepository = GroupRepositoryImpl()): ViewModel() { /* ... */ }
```

`IGroupRepository` can be the actual implementation (default parameter) or a mock that implements `IGroupRepository` for your tests. This is better because `GroupsViewModel` likely does not need to know about

RaulinN: Another comment regarding the db (again, for future projects): leverage `suspend functions` instead of using callbacks in some places. This improves readability, error handling, and consistency with modern kotlin practices

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -34,8 +34,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.tasks.await
 import kotlinx.coroutines.withContext
 
-class DatabaseConnection {
-
+class DatabaseConnection : DbRepository{
```
RaulinN, line 37: :+1:

app/src/main/java/com/github/se/studybuddies/ui/groups/GroupScreen.kt
```diff
@@ -69,7 +70,8 @@ fun GroupScreen(
     groupUID: String,
     groupViewModel: GroupViewModel,
     chatViewModel: ChatViewModel,
-    navigationActions: NavigationActions
+    navigationActions: NavigationActions,
+    db: DbRepository
```
RaulinN, line 74: Note: another common practice is to have default arguments for db manual injection, e.g.

```kotlin
db: DbRepository = DatabaseConnection.getInstance() // using the singleton design pattern
```

This way, you don't have to pass the db to every component like you did in this PR (=> more convenient)

app/src/main/java/com/github/se/studybuddies/viewModels/DirectMessageViewModel.kt
```diff
@@ -3,15 +3,14 @@ package com.github.se.studybuddies.viewModels
...
+import com.github.se.studybuddies.database.DbRepository
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 
-class DirectMessageViewModel(userUid: String = "") : ViewModel() {
-  private val db = DatabaseConnection()
+class DirectMessageViewModel(private val userUid: String = "", private val db: DbRepository) : ViewModel() {
```
RaulinN, line 13: I am surprised this even compiles! :o I did not know you could define mandatory arguments *after* optional arguments in kotlin

RaulinN (commented):
Sneaking in a few comments while the PR is still in dev =)

app/src/main/java/com/github/se/studybuddies/ui/groups/GroupScreen.kt
Ovoland: Thank you so much for your comment ! That's a small adjustment that can improve the code thanks !

app/src/main/java/com/github/se/studybuddies/viewModels/DirectMessageViewModel.kt
Ovoland: Haha same I wasn't sure about it at all at the beggining

app/src/main/java/com/github/se/studybuddies/database/DbRepository.kt
Ovoland: Thanks for your clear and very relevant comment. When I was reading documentation on FakeDatababase, I also realized that we could have done a better database implementation. But indeed I guess it's for a future project :)

TdVBd (approved):
Great coode overall, the size of the PR is not an actual problem since most lines are a copy of the original database and quick to read through
Small comment, the way you implemented the mock topics is quite different from the original ones, there might be room for improvment there 
Also a lot of tests are commented, it would be nice to solve these next

Ovoland:
This PR is really big but most of the lines are just copy pasting of the Database connection. It was hard to me to split this in two, as DbRepository needs all the functions from the Database connection to work (same for the MockDatabase). 
Also even after implementing the fake database, I still have the same issue when running the tests on git ;-; 
Thus there're multiple tests that still need to be fixed to assess that the fake database is correctly set up. Also I think that I don't have implemented perfectly the topics functions in the mockdatabase and I should probably refactor this.

PR #214, opened 2024-05-17
correction shared timer 

hanelmouta:
In this pr I tried to correct the bug that we had with the shared timer where the timer didn't start for user of the same group. I changed the way the timer value was fetched and stored in the database. I also implemented a way to make the timer synchronise for all the user. The shared timer runs in the background without having a unique instance of the shared timer like before. I deleted the use of singleton. 
Ps : There will be another or that correct some issue with the code implementation and this particular issues  **Concerning the shared timer, the changes made by an other user aren't displayed if we are already on the sharedtimer screen, but only when reloading the page.**


app/src/main/java/com/github/se/studybuddies/ui/timer/TimerScreenContent.kt
cemuelle, line 1: Nice to have set the buttons (start, pause, reset) the same positions for the solotimer and the shared one

app/src/main/java/com/github/se/studybuddies/viewModels/TimerViewModel.kt
cemuelle, line 1: Nice the fix to prevent weird behaviour when spamming the start button

cemuelle (changes requested):
You make some modifications that are welcome like have the same button position between solo and shared timer or make the shared timer working. However there is still some unwanted behaviour.
- You well fix the bug when spamming the start button in the solo timer, however you forgot to do the same for the shared timer.
- Concerning the shared timer, the changes made by an other user aren't displayed if we are already on the sharedtimer screen, but only when reloading the page. It's a big issue that need to be fixed !

PS : I found that that the timer weren't exactly the same between my phone and the emulator (few seconds of difference), it's maybe caused because they're clock could be not accurately synchronized (so you cannot do anything)

app/src/main/java/com/github/se/studybuddies/viewModels/SharedTimerViewModel.kt
```diff
@@ -1,64 +1,55 @@
...
-  init {
-    subscribeToTimerUpdates()
+  private fun startPeriodicSync() {
+    viewModelScope.launch {
+      while (isActive) {
+        syncTimerWithFirebase()
+        delay(2000) // Appeler toutes les 2 secondes
+      }
+    }
   }
```
cemuelle, line 36: It's not the more elegant way. Look at PR #217 where with the changes, it detect automatically the update and doesn't need to call it every 2s

hanelmouta: I will address this issue in another pr where I correct a lot of non good   practice of coding

cemuelle:
I cannot test because we exceeded the quota of the firebase

app/src/main/java/com/github/se/studybuddies/viewModels/SharedTimerViewModel.kt
hanelmouta: there will be another pr that address this issues

cemuelle (approved):
Still buggy, however it's working.
The optimization could be done in a new PR as we want to merge it before end of M3

PR #215, opened 2024-05-17
VideoCall Debugging

Freezemoon:
- Solve returning to call crash
- Add Toast when permissions were denied

Bugs still remaining : 
- Camera in Video Call isn't displayed


app/src/main/java/com/github/se/studybuddies/ui/video_call/CallLobbyScreen.kt
```diff
@@ -41,8 +46,24 @@ fun CallLobbyScreen(
 ) {
   LockScreenOrientation(orientation = Configuration.ORIENTATION_PORTRAIT)
```
cemuelle, line 47: I think you solved the source of the bug and we didn't need anymore the patch to prevent the rotation of the screen !

app/src/main/java/com/github/se/studybuddies/ui/video_call/VideoCallScreen.kt
```diff
@@ -1,44 +1,90 @@
...
 @Composable
-fun VideoCallScreen(call: Call, videoVM: VideoCallViewModel, onCallDisconnected: () -> Unit = {}) {
+fun VideoCallScreen(call: Call, onCallDisconnected: () -> Unit = {}) {
   val isCameraEnabled by call.camera.isEnabled.collectAsState()
   val isMicrophoneEnabled by call.microphone.isEnabled.collectAsState()
   val layout by remember { mutableStateOf(LayoutType.DYNAMIC) }
+  val connection by call.state.connection.collectAsStateWithLifecycle()
+  val context = LocalContext.current
 
   LockScreenOrientation(orientation = Configuration.ORIENTATION_PORTRAIT)
```
cemuelle, line 48: Same here, as the bug has been resolve, we can allow again to rotate the screen !

cemuelle:
I test and I figure that as you said the camera isn't working in the call, but neither the mic.

cemuelle:
Nice checks of the permissions if they are not granted with the toast and prevent to access to the video call.

app/src/androidTest/java/com/github/se/studybuddies/tests/VideoCallTest.kt
cemuelle, line 1: You should better delete completely the test than just comment everything

app/src/main/java/com/github/se/studybuddies/MainActivity.kt
cemuelle, line 1: Good to extract to a function rather than putting everything directly in the main. It should help with the complexity !

app/src/main/java/com/github/se/studybuddies/ui/video_call/CallLobbyScreen.kt
```diff
@@ -41,8 +46,24 @@ fun CallLobbyScreen(
...
+  val isLoading by callLobbyViewModel.isLoading.collectAsState()
   val isCameraEnabled by call.camera.isEnabled.collectAsState()
   val isMicrophoneEnabled by call.microphone.isEnabled.collectAsState()
+  val context = LocalContext.current
+
+  LaunchCallPermissions(
+      call = call,
+      onPermissionsResult = {
+        if (it.values.contains(false)) {
+          Toast.makeText(
```
cemuelle, line 58: I like that you displayed a Toast if the permissions are not granted, but is it possible to ask again because if it was a missclick the user should go to the android settings to manually allow the permissions to be able to use the videocall function

app/src/main/java/com/github/se/studybuddies/ui/video_call/StreamVideoHelper.kt
```diff
@@ -68,17 +69,16 @@ object StreamVideoInitHelper {
                         id = loggedInUser.uid,
                         name = username,
                     ),
-                token =
-                    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiSm9ydXVzX0NfQmFvdGgiLCJpc3MiOiJodHRwczovL3Byb250by5nZXRzdHJlYW0uaW8iLCJzdWIiOiJ1c2VyL0pvcnV1c19DX0Jhb3RoIiwiaWF0IjoxNzE0NjUzOTg0LCJleHAiOjE3MTUyNTg3ODl9.WkUHrFvbIdfjqKIcxi4FQB6GmQB1q0uyQEAfJ61P_g0",
+                token = StreamVideo.devToken(loggedInUser.uid),
```
cemuelle, line 72: Nice to remove the token written in clear !

app/src/main/java/com/github/se/studybuddies/ui/video_call/VideoCallScreen.kt
```diff
@@ -1,44 +1,90 @@
...
+        if (StreamVideo.instance().state.activeCall.value == call) {
+          Log.d("MyPrint", "Active call is the same as the call we are trying to join")
+        } else {
+          try {
+            call.join()
+            Log.d("MyPrint", "Trying to join call")
+          } catch (e: Exception) {
+            Log.d("MyPrint", "Trying to join call got exception, leave call")
+            // call.leave()
+            // call.join()
```
cemuelle, line 79: Could we display in the log what exception we good and remove the 2 commented lines (as we not really join the call, so we cannot leave it, and we don't want to join if there is something that didn't went well)

cemuelle:
I managed to make the app crash when using the android back button (the triangle that look on the left) (or gesture, when we slide from outside of the screen )

cemuelle (commented):
Good job, we have a big issue that the camera and the mic aren't working.
And also the android back make the app crash.
However the app is way more stable than before ! And we can also remove the limitation that prevent the rotation of the screen

app/src/main/java/com/github/se/studybuddies/ui/video_call/CallLobbyScreen.kt
Freezemoon: If u go to the call lobby again after denying the permissions, it should ask it again. Thats what happened when I tested it at least

app/src/androidTest/java/com/github/se/studybuddies/tests/VideoCallTest.kt
Freezemoon: Its to debug the tests in the next PR cause it's still not perfect and maybe we will have found the problem by then.

app/src/main/java/com/github/se/studybuddies/ui/video_call/VideoCallScreen.kt
Freezemoon: It doesnt change anything.

cemuelle (approved):

PR #216, opened 2024-05-18
Reduce Complexity

cemuelle:
Modify chatScreen to reduce complexity and also add more test for chatScreen and DirectMessage


app/src/androidTest/java/com/github/se/studybuddies/tests/ChatTest.kt
TdVBd, line 1: Its nice that you managed to fit all the tests in a single class instead of creating a new class for each composable like you used to do

app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
TdVBd, line 1: replacing the different AlertDialogs conditional calls with the ShowAlertDialog funtion is a smart way to reduce cognitive complexity, and makes the code clearer overall

TdVBd (approved):
Great code overall, test coverage is satisfying and cognitive complexity reduced as expected 
Maybe split this into two PRs next time, altough most lines added were just slight modifications of old code so it was not as exhausting to review as the size suggests

PR #217, opened 2024-05-19
Display automatically new Topics

cemuelle:
I fix in the group screen the issue where new topics weren't automatically displayed if we weren't reloading the page.
So when we add a new topics, it's now displayed.
It's also displaying the new topic if it's someone else that add it and we are already on the groupScreen


Freezemoon (approved):
Looks good ! Good job solving this issue ! The code is clean and there's nothing much I can find that can be improved. Nice !

PR #219, opened 2024-05-19
Optimize Database

cemuelle:
Optimize DatabaseConnection that become very big to separate the realtime and the storage part

### DIFF
--- a/app/src/main/java/com/github/se/studybuddies/database/RealtimeDatabaseConnection.kt
+++ b/app/src/main/java/com/github/se/studybuddies/database/RealtimeDatabaseConnection.kt
@@ -12,12 +12,12 @@ import com.google.firebase.database.DatabaseError
 import com.google.firebase.database.ValueEventListener
 import com.google.firebase.database.ktx.database
 import com.google.firebase.ktx.Firebase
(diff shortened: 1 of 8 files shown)


hanelmouta (commented):
Hi , good job on trying to reduce the complexity of the database class that was becoming really huge. I saw that you only modified few of the view model what about the that used DatabaeConnection I tried the app and everything seems to work correctly

app/src/main/java/com/github/se/studybuddies/database/RealtimeDatabaseConnection.kt
```diff
@@ -0,0 +1,299 @@
...
+        messageData[MessageVal.TYPE] = MessageVal.LINK
+        saveMessage(messagePath, messageData)
+      }
+      else -> {
+        Log.d("MyPrint", "Message type not recognized")
+      }
+    }
+  }
+
+  fun getMessages(
```
hanelmouta, line 118: it is a small critique you can maybe change the name of the function so that it is clear what they do for example name this one fetchChatMessages

hanelmouta:
make sure to have the good format

PR #220, opened 2024-05-19
Reduce complexity and Save Button

cemuelle:
- Optimizing the code in order to reduce the complexity
- Adding a save button for Photo and File messages in order to store them on the android device itself


Freezemoon (commented):
Very good code quality overall. Welcomed modularization of very long code to make them easier to understand and as such reduce the complexity. I wrote some comments down in parts where that wasn't too clear for me but otherwise, nice job !

Freezemoon (changes requested):
Very good code quality overall. Welcomed modularization of very long code to make them easier to understand and as such reduce the complexity. I wrote some comments down in parts where that wasn't too clear for me but otherwise, nice job !

PR #221, opened 2024-05-19
add_api_key

cemuelle:
I know that it is a bad practise, however the maps doesn't work when generated with automatic APK


hanelmouta (approved):
all good

PR #231, opened 2024-05-19
CreateToDo UI update

TdVBd:
Added the base structure to implement tests to the todo list
Fixed a bug causing the due date of tasks to not get saved correctly
Updated the UI of the Edit and Create ToDo Screens
Updated the UI of the TodoList screen
Updated the tasks status and its usages to let the user modify it more conveniently
Other small improvements, fixes and code reformating

### DIFF
--- a/app/src/main/java/com/github/se/studybuddies/ui/todo/EditToDo.kt
+++ b/app/src/main/java/com/github/se/studybuddies/ui/todo/EditToDo.kt
@@ -4,7 +4,6 @@ package com.github.se.studybuddies.ui.todo
 //noinspection UsingMaterialAndMaterial3Libraries
 import android.annotation.SuppressLint
 import android.util.Log
-import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/github/se/studybuddies/ui/todo/EditToDo.kt
```diff
@@ -75,87 +80,66 @@ fun EditToDoScreen(
     val isOpen = remember { mutableStateOf(false) }
     val expanded = remember { mutableStateOf(false) }
```
oliver216, line 81: Val expanded is never use, you can safely remove it

oliver216 (approved):
Very good code, clean and understandable. Good PR size.

Nothing to say about the code directly, the UI is correctly implemented. The test is not finished. And the function is compact and clean. Add just added one small comment on one line.

I tested the app on my computer, the date on the task now saves correctly, well done !
The color blue you use for the button is still not the one use in the rest of the app, could be great to have the same everywhere (one si Color.Blue, other is just Blue). Also, when selecting a date the selected one become blue and the number selected is not visible anymore. 
A lot of texts are in gray, some could be in blue for better view, like the title "select date" so that it is more visible. (Some are good in gray and should not be changed like "title" and "description" when they are not selected, it gives a good view of where I am writing.)
Create, Edit, Save works great ! Very small detail, when a task is marked as "done" the text is a little more in the center which create a little gap from the others task.

The start of the test is good, coverage is too low but logical as the test is not finished.

Overall a good work !

PR #236, opened 2024-05-22
Fixing GroupHomeTest with MockDatabase

Ovoland:
I'm correcting the mock database for the group home tests. I'm also improving these tests to verify more aspects of the screen, such as the group option button
For example, I check that the option of entering correct and wrong links work fine. By doing this I spotted that there was a mistake with this feature. When the user doesn't have any group, the link textfield was taking too much place (the box was too large) and couldn't be recognized by the tests.  Also I found issue when entering wrong links. Probably due to conflict during merge, the code checking for correct links was removed and we were accessing bugged groups when entering wrong links. 
I fixed both of theses issues. I also changed the way of handling the waiting process before loading. I used the same technique as for the map as it has the advantage of stopping the loading as soon as the data are fetched.
Plus I also corrected few things with the mockdatabase, such as the topics which were wrongly implemented.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,11 +1,9 @@
 name: CI - Test Runner
 
-# Run the workflow when commits are pushed on main or when a PR is modified
 on:
   push:
     branches:
       - main
-
   pull_request:
     types:
       - opened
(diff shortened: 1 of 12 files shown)


Ovoland:
I had a lot of troubles finding the group items of the LazyColumn and I lost a lot of time before finding that we could find element using
   composeTestRule
          .onNodeWithTag("groupTest1_settings_button", useUnmergedTree = true)
This helps me a lot to find UI element with their tags

Ovoland:
My new main issue is that the tests are (one more time) passing perfectly fine locally but failed on git :/ Some UI elements aren't recognize and I have no clue why :/ 
I have tried the tests on my tablet and on the pc emulator (API 34) but everything seems correct locally ;-;

TdVBd (approved):
Great job overall, code is clean, the tests are functional and implemented uniformly
I see some are still commented, whould be nice to get them working when possible

PR #244, opened 2024-05-24
Video call debugging : Part 1 

Freezemoon:
Debugged issues : 
- Video and microphone work fine in the call now
- Turning the screen won't cause crashes

Remaining issues : 
- Video and microphone not yet fully synchronized when first joining a call 
- Rotating the screen makes some buttons disappear, add scrolling ability in case of rotation 
- Leaving a call still doesn't actually leaves the call, there's a notification saying there's still a call in progress while supposed to have left it


Ovoland (approved):
Nice fixes that help the video call feature a lot ! The issues you mentioned first seem to be definitively fixed.
Your video call code is quite complex and I don't really know how to help you for the remaining bugs... 
Also I notice that after leaving the video call, the call indeed doesn't stop and the app crash if you try to leave the group (or do other actions). 
Good luck for the rest of the debugging and well done for these fixes !

PR #246, opened 2024-05-24
Add group member from list

oliver216:
In group, to add new members, add a list of all "friend" (for now all users). When clicking on one, it adds it to the group, or you can exit this page by clicking on the cross

Add a button to show this page on GroupSetting (Now there are two ways to add a member to a group)

Add the two buttons to add a member to GroupMember Page

Clean some code that called the DataBase for each member of a group and replace it by only one call

Correct a few bug and UI problem

### DIFF
--- a/app/src/main/java/com/github/se/studybuddies/viewModels/GroupViewModel.kt
+++ b/app/src/main/java/com/github/se/studybuddies/viewModels/GroupViewModel.kt
@@ -79,8 +79,8 @@ class GroupViewModel(uid: String? = null, private val db: DbRepository = Databas
     viewModelScope.launch { db.deleteGroup(groupUID) }
   }
 
-  fun addUserToGroup(groupUID: String, text: String = "") {
-    viewModelScope.launch { db.addUserToGroup(groupUID, text) }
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/se/studybuddies/viewModels/GroupViewModel.kt
```diff
@@ -72,15 +88,42 @@ class GroupViewModel(uid: String? = null, private val db: DbRepository = Databas
...
+      db.addUserToGroup(groupUID, text) { isError -> callBack(isError) }
+      val userUID: String =
+          if (text != "") {
+            text
+          } else {
+            db.getCurrentUser().toString()
+          }
+      val newUser = db.getUser(userUID)
+      var updatedMembers = _members.value?.toMutableList()
+      if (newUser != null) {
```
hanelmouta, line 121: maybe add a callback here callBack(true)

hanelmouta (commented):
I tested the code and it looks good. The implementation of the addUserToGroup function now includes a callback to handle success or failure, and logs appropriate messages. The new implementation is integrated into the GroupViewModel and UI components.

hanelmouta (approved):
good job I tested and everything works fine

PR #247, opened 2024-05-24
Calender 

hanelmouta:
you can now delete goals appointment  or note

### DIFF
--- a/app/src/main/java/com/github/se/studybuddies/MainActivity.kt
+++ b/app/src/main/java/com/github/se/studybuddies/MainActivity.kt
@@ -52,6 +52,7 @@ import com.github.se.studybuddies.ui.video_call.CallLobbyScreen
 import com.github.se.studybuddies.ui.video_call.StreamVideoInitHelper
 import com.github.se.studybuddies.ui.video_call.VideoCallScreen
 import com.github.se.studybuddies.viewModels.CalendarViewModel
+import com.github.se.studybuddies.viewModels.CalendarViewModelFactory
 import com.github.se.studybuddies.viewModels.CallLobbyViewModel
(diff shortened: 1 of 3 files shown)


Ovoland (approved):
Nice improvement of the calendar ! The functionality seems to work nicely :) Plus you used the viewModelFactory as the TAs asked which is really nice ! 
For me, the code seems mostly good to me and I mostly only have UI recommendations. 
But first I noticed some bugs, for example if you create a notes/goal/appointement and then rotate the screen, the added thing will diapear. 
Then I  have some suggestions concerning the UI which could maybe a bit hard to understand for some users. For example, to delete or edit a note/goal/appointement, I would try to use the same method as for the groups deletion (click on 3 dots button to open a small window and chose delete/edit). Thus in my opinion it would be a bit more intuitive for the user and it would fit the general deleting/editing mode. 
Also if you want to change the UI, I have some suggestions concerning other UI element (the timer selection in the Appointements, the notes visible on the "calendar home", way of saving the planner) but I think it's easier if we can just talk about together by call or in real. 
Finally for a later update, it would probably better to change the way the items daily planner are stored in the database. Because if you take a look at the database, you'll see that these items are stored next to the users personal/important informations. 
image
I'm a bit afraid that storing daily planner item there will overfload the essential userData. It would probably better to do something similar to the way we store groups. You could probably create a collection for Daily Planner items. 
In conclusion, this PR is good and brings a new useful features :) If you have time, for later PR, I think you could try to see how to improve some UI aspects and we talk about this if you wish
But for now, the code is ready to be merged

hanelmouta:
thank you for your review I will try to address the changes asked in another pr

PR #248, opened 2024-05-24
Strong users

Svannia:
- There is a switch on topic files to select oneself as a strong user
- new viewModel for topic files
- new screen for data related to topic files (resources to be implemented, and list of strong users)
- UI re-adjustements


Ovoland (approved):
Overall very good PR ! The feature seems to work really nicely !
I only have few comments on minor part. First when we reload the screen, add a new file, the toggle buttons go from unchecked to checked (for the file with strong user) rapidly. It's a bit confusing, especially when adding new files. Indeed new added files are placed depending on the alphabetic order. Thus if you have a file with strong user and that you add another file that is placed above the first, the file with strong user will be reload and the toggle will move again, which is a bit confusing.
Maybe it would be interesting to allow the page to load for few seconds before rendering the screen. Thus the files would have the time to get the information concerning strong users. 
Also as mentioned with the TAs, we should stop initiating Viewmodel.

Svannia:
> Overall very good PR ! The feature seems to work really nicely ! I only have few comments on minor part. First when we reload the screen, add a new file, the toggle buttons go from unchecked to checked (for the file with strong user) rapidly. It's a bit confusing, especially when adding new files. Indeed new added files are placed depending on the alphabetic order. Thus if you have a file with strong user and that you add another file that is placed above the first, the file with strong user will be reload and the toggle will move again, which is a bit confusing. Maybe it would be interesting to allow the page to load for few seconds before rendering the screen. Thus the files would have the time to get the information concerning strong users. Also as mentioned with the TAs, we should stop initiating Viewmodel.

Hi, thanks for the feedback! I added a loading on fetch, however I could not get rid of the switch sliding to its checked position, because each file is fetching their own data and unless I put on a "loading" animation for each file, I'm not sure how to do this... I'll keep this in mind though

PR #249, opened 2024-05-25
CreateToDo tests

TdVBd:


app/src/androidTest/java/com/github/se/studybuddies/tests/CreateToDoTest.kt
```diff
@@ -32,80 +39,199 @@ class CreateToDoTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withCompose
...
-            // arrange: verify pre-conditions
-            assertIsDisplayed()
-            performClick()
-          }
-        }
-        // assert: the nav action has been called
-        verify { mockNavActions.navigateTo(Route.TODOLIST) }
-        confirmVerified(mockNavActions)
+  @Test
+  fun inputDate() {
```
Freezemoon, line 141: Not sure if you forgot to delete this commented out code or if the test isn't passing the CI yet ?

app/src/main/java/com/github/se/studybuddies/ui/shared_elements/SharedUI.kt
```diff
@@ -365,6 +339,29 @@ private fun MenuButton(onClick: () -> Unit) {
...
+          )) {
+        Text(
+            stringResource(R.string.save),
+            color = White,
+            modifier = Modifier.testTag("save_button_text"))
+      }
+}
+
+/*
 @Composable
```
Freezemoon, line 365: The chunk of commented out code could probably be deleted

app/src/main/java/com/github/se/studybuddies/viewModels/ToDoListViewModel.kt
```diff
@@ -33,9 +33,7 @@ class ToDoListViewModel(studyBuddies: Application) : AndroidViewModel(studyBuddi
   private val json = Json { prettyPrint = true }
   // private val gson = Gson()
 
-  private val toDoFile = File(studyBuddies.filesDir, "ToDoList.json")
-
-  // val encryption = Encryption(studyBuddies)
+  private val toDoFile = File(context.filesDir, "ToDoList.json")
 
   init {
```
Freezemoon, line 38: Just in case, the coaches sent a very useful link about viewmodels I think two weeks ago and it said that using init is bad practice for viewmodels as they could cause problems in the tests (though it seems to be working fine here) so just a head's up !

Freezemoon (commented):
Good job ! Very nice that you managed to better modularize some UI parts to allow better understanding and comprehension of the code. Also minimizes the complexity and more intuitive use of components !

app/src/main/java/com/github/se/studybuddies/viewModels/ToDoListViewModel.kt
TdVBd: Okay thank you, I will check that, might have time to update it if useful

app/src/main/java/com/github/se/studybuddies/ui/shared_elements/SharedUI.kt
TdVBd: Done!

app/src/androidTest/java/com/github/se/studybuddies/tests/CreateToDoTest.kt
TdVBd: It has not passed CI yet, i will try to solve this later if given the time

Freezemoon (approved):
Nice, good job!

PR #256, opened 2024-05-27
GroupScreen and GroupSetting tests

Ovoland:
I have implement tests for the groupScreen and GroupSetting


Ovoland:
The topic items were detected locally but not on git. Despite trying a lot of different approches to solve this I didn't managed to find the topic items. Thus I skip the topic verification on the testing file. 
Concerning the group setting screen I choose to stop at that point because the PR was already too big. The rest of the tests need to be implemented later.

app/src/main/java/com/github/se/studybuddies/ui/groups/GroupScreen.kt
```diff
@@ -186,19 +197,20 @@ fun TopicItem(groupUID: String, topic: Topic, navigationActions: NavigationActio
...
-        Row(modifier = Modifier.fillMaxWidth().padding(16.dp)) {
-          Spacer(modifier = Modifier.size(16.dp))
+          Modifier.fillMaxWidth().testTag("${topic.uid}_item").background(Color.White).clickable {
+            navigationActions.navigateTo("${Route.TOPIC}/$topicUid/$groupUID")
+          }
+      /*.drawBehind {
+        val strokeWidth = 1f
+        val y = size.height - strokeWidth / 2
+        drawLine(Color.LightGray, Offset(0f, y), Offset(size.width, y), strokeWidth)
+      }*/
```
cemuelle, line 207: If useless, we should remove it

cemuelle (approved):
Nice to have implemented test for the GroupScreen and GroupSetting that weren't have any!
Clean tests.
Just in GroupScreen you commented some lines but if they are useless you should delete them.

app/src/main/java/com/github/se/studybuddies/ui/groups/GroupScreen.kt
Ovoland: True ! I will uncomment this part !

PR #259, opened 2024-05-28
Delete topic

Svannia:
Delete button in the topic settings that:
- Fetches the topic in the database
- For both exercises and theory lists of topic items, fetches them
- For each topic item
     -> if it is a file, delete it
     -> if it is a folder, iteratively delete its content and then delete the empty folder
- delete the empty topic


Freezemoon (changes requested):
The code is very clean and efficient. Very nice to iterate over the topics using recursion, it's a simple yet very efficient solution. The size of the PR is just right and the implemented functions and UI components are all nicely uniformized with the current project UI and strucutre. One thing I have to add, it would be nice to incorporate a popup when a user wants to delete a topic to confirm their choice. We wouldn't want to accidentally delete a useful topic because of a missclick ! This is the only thing I would suggest to add as I know it could happen to me... Otherwise, good job !

Freezemoon (approved):

PR #263, opened 2024-05-29
Filter messages

cemuelle:
Be able to search a message with his content or to filter by type of message


oliver216 (changes requested):
Really a good work ! The code is clean and well written.

The filter works great, are quick and intuitive ! Some small problem :
The color of the search bar (grey) doesn't really fit with the rest of the page/app, a lighter color could be better.
When closing the search bar by clicking on it again, the text input for searching is not reset thus the messages are still filtered, but if for example the filter "image" was chosen, when closing the filter is restart to "All". It's not consistent, either all filter reset when closing or nothing.
Lastly, when the "Link" filter is pressed I can't press "File" and reciprocally. The other filter can be pressed correctly.

The test coverage is not very high.

Overall a good job !

cemuelle:
Nice catch of the bugs,:
- I put the searchBar in white
- Apply the same behaviour of remembering the filter that was set
- update the UI of the button that has been selected even if nothing match it

And I complete a little bit the tests
And also try to fix an sonar issue

oliver216:
Good correction, works fine ! And the test coverage increased a lot, great !

oliver216 (approved):

PR #269, opened 2024-05-31
Correction timer

hanelmouta:
I correct the implemtation of the shared timer now any user of the group can start the timer it will start for everyone pause or reset


app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -341,41 +342,29 @@ class DatabaseConnection : DbRepository {
...
 
-    // Create a map for the new timer state
-    val newTimerState = mapOf("endTime" to newEndTime, "isRunning" to newIsRunning)
-
-    // Update the timerState field in the group document
     try {
       groupDataCollection
           .document(groupUID)
-          .update("timerState", newTimerState)
+          .update("timerState", timerState)
```
cemuelle, line 354: You should have a const timerState instead of the string in clear

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -389,7 +378,7 @@ class DatabaseConnection : DbRepository {
       val picture = Uri.parse(document.getString("picture") ?: "")
       val members = document.get("members") as List<String>
       val timerStateMap = document.get("timerState") as? Map<String, Any>
-      val endTime = timerStateMap?.get("endTime") as? Long ?: System.currentTimeMillis()
+      val endTime = timerStateMap?.get("endTime") as? Long ?: 0L
       val isRunning = timerStateMap?.get("isRunning") as? Boolean ?: false
```
cemuelle, line 382: Again it's will be great to have const, so you could reuse the timerState defined in updateGroupTimer

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -1293,25 +1282,37 @@ class DatabaseConnection : DbRepository {
...
-            override fun onCancelled(error: DatabaseError) {
-              Log.e("TimerViewModel", "Failed to read timer", error.toException())
-            }
-          })
-    } ?: error("Group UID is not set. Call setup() with valid Group UID.")
-    return isRunning
+      if (snapshot != null && snapshot.exists()) {
+        val timerStateMap = snapshot.get("timerState") as? Map<String, Any>
+        val endTime = timerStateMap?.get("endTime") as? Long ?: 0L
+        val isRunning = timerStateMap?.get("running") as? Boolean ?: false
```
cemuelle, line 1302: And reuse the const here

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
```diff
@@ -1293,25 +1282,37 @@ class DatabaseConnection : DbRepository {
...
+        val timerState = TimerState(endTime, isRunning)
+        CoroutineScope(ioDispatcher).launch {
+          val remainingTime = endTime
+          withContext(mainDispatcher) {
+            _timerValue.value = remainingTime
```
cemuelle, line 1307: You could inlined this, remainingTime is set to endTime and then _timerValue.value use it, but only there. So you could remove it and directly set `_timerValue.value = endTime`

cemuelle (changes requested):
Nice code,
You just need to put the strings into a const and inlined a variable.

When testing, sometimes it work well.
However when accessing from an old group i get a weird number on the screen.
And with a newly created group, if we click multiples times on start it "blocks" the countdown. Additionnaly I managed to "broke" it and on one device the countdown isn't working, but another yes (even reinstall the app doesn't solve the issue. But's it maybe due to the quota that is full.
So you can make sure that pressing a button when we shouldn't doesn't do anything (as pressing on start when already running, on pause when stop, ...)

cemuelle:
Great job, the shared timer is working well.
Just you should add a check if the timer is already running and a user press on the start button, because now it's blocking if an user click multiple times on it (and prolonged a little bit the timer if someone click just once).

Additionnaly if an user add some time and another click on reset at the same, we got an weird behaviour even after we stop touching any button (it's working well when we click multiples times in a small period). The digit are blinking

https://github.com/Study-Buddies-SwEnt/SwEnt_Group_project/assets/55513917/c6fee2cb-2942-490e-b71d-1ee9e89ac93e

PR #270, opened 2024-05-31
Group test bugs UI

oliver216:
Add "add member" option in GroupHome to directly add a member to a group from there
Add a loading text when waiting on the contact list to load

Correct some big bugs that had appeard : Correct a crash when leaving a group. 
Correct a crash when entering an empty link to join a group.
Correct a crash when entering a correct link.
Correct that when entering a wrong userUID to add a member, it doesn't create anymore an empty user (was only visual, not on the database)

Add some tests for these new functions and some old one.


app/src/main/java/com/github/se/studybuddies/ui/groups/MembersList.kt
```diff
@@ -0,0 +1,56 @@
...
+  if (groupUID.isEmpty()) return
+  val isBoxVisible = remember { mutableStateOf(true) }
+
+  Scaffold(
+      modifier = Modifier.fillMaxSize().background(White),
+      topBar = {
+        TopNavigationBar(
+            title = { Sub_title(stringResource(R.string.add_member)) },
+            navigationIcon = {
+              GoBackRouteButton(navigationActions = navigationActions, Route.GROUPSHOME)
```
Svannia, line 42: I realized that on the 5 different screens of group settings (modify group, ..., delete group) when pressing the back button we navigate to GroupsHome, which feels a bit weird. Instead of this line, can you just call `navigationActions.goBack` (for the 5 screens)? This way the app will just navigate to the previous screen and it will feel much better.

app/src/main/java/com/github/se/studybuddies/ui/groups/GroupMembers.kt
```diff
@@ -269,3 +273,96 @@ fun MemberOptionButton(
...
+        item { Spacer(modifier = Modifier.height(64.dp)) }
+        item {
+          Box(modifier = Modifier.fillMaxSize()) {
+            IconButton(
+                onClick = { isBoxVisible.value = false },
+                modifier = Modifier.align(Alignment.TopEnd).testTag("close_contact_list")) {
+                  Icon(
+                      imageVector = Icons.Default.Close,
+                      modifier = Modifier.size(40.dp),
+                      contentDescription = "Close friends List",
```
Svannia, line 313: I don't think this icon is necessary, since it does the same thing as the go_back button. Unless it has a purpose I misunderstood, I believe you can remove it

Svannia (changes requested):

Svannia (changes requested):
I don't know if something unexpected changed when you pushed, but on your branch, tapping the button to enter a link to join a group doesn't work: there's nothing happening. Maybe it's just that the textField isn't visible?

app/src/main/java/com/github/se/studybuddies/ui/groups/GroupMembers.kt
oliver216: It is useful when there is no goBack button, which is not the case in this situation but at other place yes. And I don't think it's bad to have the cross to cancel the add friend menu even if you can use goBack.

oliver216:
I don't have this issue on my side, maybe you have too much groups and you need to scroll to see the Enter Link input ?

app/src/main/java/com/github/se/studybuddies/ui/groups/MembersList.kt
oliver216: Yes you are correct, I change the function used for the group so that it goes back to last screen.

Svannia (approved):

Svannia:
Okay! Thanks for changing the back navigation.
For the join-group link not appearing for me, it was in fact below the list of groups. This isn't relevant to this PR, but it's true that this textField as well as the floating buttons should have been on top of the screen, not below the list of groups

PR #271, opened 2024-06-01
Chat poll

cemuelle:
Add poll in chat :
- Create new poll
- Display the poll in the chat
- Vote in the poll
- Modify his vote
- display the number of votes for each options
- enable the filtering by type for Poll

- Fix edit message that was not anymore working


app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
```diff
@@ -927,6 +990,78 @@ fun FilePickerBox(
...
+          Row(
+              modifier =
+                  Modifier.fillMaxWidth().padding(8.dp).testTag("add_poll_single_choice_row"),
+              verticalAlignment = Alignment.CenterVertically,
+              horizontalArrangement = Arrangement.SpaceBetween) {
+                Text(
+                    text = stringResource(R.string.single_choice),
+                    style = TextStyle(color = Black),
+                    modifier = Modifier.testTag("add_poll_single_choice_text"))
+                Switch(checked = singleChoice.value, onCheckedChange = { singleChoice.value = it })
```
Svannia, line 1048: Could you change the default colors? For coherence you can re-use the ones already set in TopicScreen. Because for now, when the switch is on, the entire track is blue and the thumb is not visible, which doesnt make it very clear that this is a switch button, and in dark mode, the unchecked switch looks kinda weird

Svannia (changes requested):
Apart from a minor UI fix, the PR looks and works great to me! Good job

app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
cemuelle: True, it was not very well display. Thanks for the indication where to find the solution!

cemuelle:
In the e2e pr, there has been an unwanted change about uploadChatFile that when sending messages it was removed.
I put it back and reduce the length of the title in directmessage

Svannia (approved):

Svannia:
Great, it looks good like that! Thanks for shortening the DMs screen title

PR #272, opened 2024-06-01
Fix main not passing CI

Svannia:
signOut test removed, it was bugged and often failed many well-build PRs


cemuelle (approved):
I also had some issue with this test.
I had to clean the cache of the Github Actions and rerun the test to make it working. So I think it's a great idea to remove the test and maybe if we have more time to fix it properly

PR #274, opened 2024-06-01
Splitting chat into multiples files

cemuelle:
To increase readibility, I split the ChatScreen that was almost 1100 lines long into multiple files with for example chatOption all the options available for the messages, SendTypeMessage for all the composable that are cold for different type of messages


TdVBd (approved):
Good job on splitting these files that were getting too big
No new code so nothing more to say

PR #275, opened 2024-06-01
Topic resources

Svannia:
In the resources tab within a topic file, the user can now add pictures.
The list of pictures is auto-updated everytime a user adds a new one.


cemuelle (changes requested):
Clean code, I like how you reuse some parts of the chat to avoid to recode and have duplication.
Just make sure to delete the parts you don't want instead of comment it up.
And it will be have great if you had move them in a shared folder. 
Last point, it's a little sad that we cannot open or save the image, because they are difficult to see with they're current size. You could reuse the download button of the chat.

app/src/main/java/com/github/se/studybuddies/ui/topics/TopicResources.kt
```diff
@@ -140,6 +181,22 @@ fun TopicResources(
...
+  Log.d("Topics", "current image is $image")
+  Row(
+      modifier = Modifier.fillMaxWidth().padding(8.dp),
+      verticalAlignment = Alignment.CenterVertically,
+      horizontalArrangement = Arrangement.Center) {
+        Image(
+            painter = rememberAsyncImagePainter(image.toString()),
+            contentDescription = stringResource(R.string.image_resource),
+            modifier =
+                Modifier.size(200.dp).clip(RoundedCornerShape(20.dp)).testTag("chat_message_image"),
```
TdVBd, line 195: is this the correct test tag or just a reused one ?

TdVBd, line 195: the image appears quite big and in the middle of my screen, making it feel like we lose a lot of useful space, is this intentional ?

app/src/main/java/com/github/se/studybuddies/ui/topics/TopicResources.kt
```diff
@@ -173,3 +230,68 @@ private fun UserBox(user: User) {
...
+) {
+  PickPicture(showUploadImage) { topicFileViewModel.addImage(it.value) }
+  // UploadLink(viewModel, showUploadLink)
+  // UploadFile(viewModel, showUploadFile)
+  ShowAlertDialog(
+      showDialog = showOptions,
+      onDismiss = { showOptions.value = false },
+      title = {},
+      content = {
+        Row(
```
TdVBd, line 250: this dialog seems quite big with the icon very small in the center, but i guess it is related to the commented code that is not displayed here yet

TdVBd (approved):
Great job overall, the code is clean, concise and the feature is behaving as expected,
I have made a few minor comments on potential improvements

Svannia:
> Clean code, I like how you reuse some parts of the chat to avoid to recode and have duplication. Just make sure to delete the parts you don't want instead of comment it up. And it will be have great if you had move them in a shared folder. Last point, it's a little sad that we cannot open or save the image, because they are difficult to see with they're current size. You could reuse the download button of the chat.

I don't think it's a good idea to download the image, because as a user you're basically kind of duplicating all the resources? But I can make it so that the image shows bigger when clicked on!

app/src/main/java/com/github/se/studybuddies/ui/topics/TopicResources.kt
```diff
@@ -173,3 +230,68 @@ private fun UserBox(user: User) {
...
+@Composable
+fun AddResources(
+    topicFileViewModel: TopicFileViewModel,
+    showOptions: MutableState<Boolean>,
+    showUploadImage: MutableState<Boolean>,
+    showUploadLink: MutableState<Boolean>,
+    showUploadFile: MutableState<Boolean>,
+) {
+  PickPicture(showUploadImage) { topicFileViewModel.addImage(it.value) }
+  // UploadLink(viewModel, showUploadLink)
```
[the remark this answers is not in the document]
Svannia: I'd like to implement it later eventually, so that more different types of resources can be added.

app/src/main/java/com/github/se/studybuddies/ui/topics/TopicResources.kt
Svannia: yep, a reused one! I missed it but it's removed now

Svannia: Mmh, I was going for a sort of "feed"-looking page, where there might later be also pdfs, links, etc...

app/src/main/java/com/github/se/studybuddies/ui/topics/TopicResources.kt
Svannia: yeah exactly. I'm leaving room for future development but I agree that it lookds a bit big now

app/src/main/java/com/github/se/studybuddies/ui/chat/ChatScreen.kt
```diff
@@ -451,3 +467,394 @@ fun PrivateChatTitle(chat: Chat) {
...
+) {
+  if (showDialog.value) {
+    AlertDialog(
+        modifier = modifier,
+        onDismissRequest = onDismiss,
+        text = content,
+        title = title,
+        confirmButton = button)
+  }
+}
```
cemuelle, line 860: Theses functions have been moved over there : https://github.com/Study-Buddies-SwEnt/SwEnt_Group_project/blob/main/app/src/main/java/com/github/se/studybuddies/ui/chat/utility/ChatOption.kt

It was to reduce the size of the file and also it's complexity.

Can you not put it back here and make you changes over there

cemuelle:
It's very cool to be able to zoom into the photo to see details, especially if we want to better see maybe a mathematical formula or something else that was shared.

If you can't change it, it's not a big deal, but can you make sure that we cannot move the image outside of the screen?

app/google-services.json
cemuelle, line 1: You shouldn't put this file in github, it should remain private

app/src/main/java/com/github/se/studybuddies/ui/chat/utility/SendTypeMessage.kt
cemuelle, line 1: I think we should keep SendPhotoMessage (or the rename version) in this file, or also move all the composable because they do all a similar job.
The other possibility is to move all of them

app/src/main/java/com/github/se/studybuddies/ui/chat/utility/Utility.kt
cemuelle, line 1: Similar Than in SendTypeMessage that we should keep PickPicture there, or move all the other functions

cemuelle (changes requested):

cemuelle (approved):
Perfect, great job !
Thanks for the correction
You have well deleted the google-service.json and by not changing the PickPicture of file, it's make more sense and your PR is more readable

cemuelle (changes requested):
Just your `fun TopicResources()` has to much complexity. Can you optimize it

cemuelle (approved):
Nice reduction of the complexity, no more sonar issue

PR #279, opened 2024-06-02
Fix link name in chat

cemuelle:
The line has mysteriously disspear in a commit, I put it back


oliver216 (approved):
Weird indeed, all good :)

PR #280, opened 2024-06-02
Group calendar

hanelmouta:
-create the calendar group where you can add


app/src/main/java/com/github/se/studybuddies/MainActivity.kt
TdVBd, line 1: its nice that you thought to use a viewModelFactory

app/src/main/java/com/github/se/studybuddies/database/DatabaseConnection.kt
TdVBd, line 1: good job on creating a new collection to store planner data, instead of adding it to the already large userData collection

app/src/main/java/com/github/se/studybuddies/viewModels/CalendarGroupViewModel.kt
TdVBd, line 1: I see a lot of code duplication that it might have been possible to avoid, through the use of an abstract class for example

TdVBd (approved):
Great job overall, it is a quite large PR because of duplicated code
Everything seems to be working as expected
only 1.8% coverage sadly, might be nice to add some easy tests if we have time

PR #281, opened 2024-06-02
Correct Shared Timer

hanelmouta:


cemuelle (changes requested):
Unfortunaly I have tested the timer on my phone and a emulator and the timer wasn't shared between them.

cemuelle:
It working again, but we still have the same issues than in PR #269 and [decribed here](https://github.com/Study-Buddies-SwEnt/SwEnt_Group_project/pull/269#issuecomment-2143882099)

app/src/main/java/com/github/se/studybuddies/viewModels/SharedTimerViewModel.kt
```diff
@@ -3,98 +3,118 @@ package com.github.se.studybuddies.viewModels
...
+  fun setTimer(hours: Long = 0, minutes: Long = 0, seconds: Long = 0) {
+    val newTime = (hours * 3600 + minutes * 60 + seconds) * 1000
+    _timerValue.value = newTime
+
+    viewModelScope.launch { db.updateGroupTimer(groupUID, TimerState(newTime, _isRunning.value)) }
+  }
+
+  fun startTimer() {
+    // Timer is already running
+
```
cemuelle, line 67: To resolve the bug when an user click multiples time on the start and it's block the app, you could simply add
`if (_running_local.value) return` to avoid the execution of the rest of the function

cemuelle:
> It working again, but we still have the same issues than in PR #269 and [decribed here](https://github.com/Study-Buddies-SwEnt/SwEnt_Group_project/pull/269#issuecomment-2143882099)

To resolve the issue of multiple clicks on Start you could just modify your code is SharedTimerScreen to add `if (!sharedTimerViewModel.isRunning.value)`
```
// Timer control buttons
          Row(
              modifier = Modifier.fillMaxWidth().padding(16.dp),
              horizontalArrangement = Arrangement.SpaceEvenly) {
                TimerButton(onClick = { if (!sharedTimerViewModel.isRunning.value) sharedTimerViewModel.startTimer() }, text = "Start")
                TimerButton(onClick = { sharedTimerViewModel.pauseTimer() }, text = "Pause")
                TimerButton(onClick = { sharedTimerViewModel.resetTimer() }, text = "Reset")
              }
```

cemuelle (approved):
Nice job to make it working.
And for the bug that occurs when spamming on one phone add time and reset on another one I think it could be a new issue/PR to resovle it.

PR #282, opened 2024-06-02
Update README.md

Freezemoon:


cemuelle (approved):
Thanks for the update of the diagram and the summary

PR #285, opened 2024-06-02
Documentation and hard-coded strings put into string.xml

Freezemoon:


Svannia (changes requested):
There are a few functions where you didn't give explanations for the parameters. Don't know if there's a reason or should someone else who knows those bits of code better do it?

Freezemoon:
> There are a few functions where you didn't give explanations for the parameters. Don't know if there's a reason or should someone else who knows those bits of code better do it?

I didnt put parameters for most functions since we kinda short on time so I only put what the function does. I did put the parameters for each class though.

Svannia (approved):

PR #288, opened 2024-06-03
Contact request

TdVBd:
Adding a contact list screen separate from direct messages to handle contacts and friend requests
Adding contact requests that need to get accepted for contact creation$
Other reformating and improvements


hanelmouta (approved):
- al good

